PR #1, opened 2013-11-23
Initial app "repos" and more project structure

txels:
Besides the app, I've started some structure for settings: `base`, `dev`, later on we will have `live` as well, and we can have our own as `dev_carles` and `dev_alex` e.g..


ployst/repos/models.py
```diff
@@ -0,0 +1,49 @@
+from django.db import models
+from south.modelsinspector import add_introspection_rules
+
+
+class Revision(models.CharField):
```
txels, line 5: I've added this custom field not just for saving us from typing the `max_length=40` but rather to later be able to attach behaviour to it.
I was considering making `Revision` a model but I don't think we will actually need that.

alexcouper:
in

ployst/repos/models.py
```diff
@@ -0,0 +1,49 @@
...
+    If required in the future, this can be extended to support other
+    repo types, such as mercurial, subversion etc.
+
+    """
+    name = models.CharField(max_length=100)
+    url = models.URLField()
+    active = models.BooleanField(default=True)
+
+    class Meta:
+        verbose_name_plural = 'repositories'
```
alexcouper, line 34: Very diligent of you ;)

alexcouper:
Looks good to me!

PR #2, opened 2013-11-24
Initial REST API endpoints for repos and branches

txels:
This sets up an initial set of ReST endpoints for repos and branches. Little work, but it is already a full-fledged API thanks to the magic of DRF. This will be the starting point for doing any initial Angular front-end.

You can browse the API (and create objects as well) at http://localhost:8000/repos/repo/

Not included in this PR are user authentication and permissions (it will be the purpose of https://txels.tpondemand.com/entity/275 and later stories, and this PR is still part of https://txels.tpondemand.com/entity/282)

### DIFF
--- a/ployst/repos/models.py
+++ b/ployst/repos/models.py
@@ -15,6 +15,10 @@ class Revision(models.CharField):
         kwargs['max_length'] = 40
         super(Revision, self).__init__(self, *args, **kwargs)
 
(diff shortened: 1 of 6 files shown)


alexcouper:
Looks really simple. Nice.

I haven't tried it out, but hopefully I'll be able to do some work on this this week.

PR #3, opened 2013-11-24
Add application 'builds' for CI jobs and build results

txels:
Following on the previous PR's footsteps, these are additional models for a `builds` app.

I think it's important that we get the data model right to begin with, then we can start building the providers and the front-end based on that.

Next step, a `features` app.

### DIFF
--- a/ployst/builds/migrations/0001_initial.py
+++ b/ployst/builds/migrations/0001_initial.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+import datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        # Adding model 'Build'
+        db.create_table(u'builds_build', (
+            (u'id', self.gf('django.db.models.fields.AutoField')(primary_key=True)),
+            ('provider', self.gf('django.db.models.fields.CharField')(max_length=40)),
+            ('job_id', self.gf('django.db.models.fields.CharField')(max_length=100)),
+            ('revision', self.gf('ployst.repos.models.Revision')(max_length=40)),
+            ('status', self.gf('django.db.models.fields.CharField')(default='gray', max_length=6)),
+            ('url', self.gf('django.db.models.fields.URLField')(max_length=200)),
+            ('build_time', self.gf('django.db.models.fields.DateTimeField')(auto_now_add=True, blank=True)),
+        ))
+        db.send_create_signal(u'builds', ['Build'])
+
+
+    def backwards(self, orm):
+        # Deleting model 'Build'
+        db.delete_table(u'builds_build')
+
+
+    models = {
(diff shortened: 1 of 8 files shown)


ployst/builds/models.py
```diff
@@ -0,0 +1,47 @@
...
+    UNSTABLE = Status('yellow', 'unstable')
+    FAILURE = Status('red', 'failure')
+    NOT_BUILT = Status('gray', 'not built')
+    ALL_STATUSES = [NOT_BUILT, FAILURE, UNSTABLE, SUCCESS]
+
+    job = models.ForeignKey(Job, related_name='builds')
+    revision = Revision()
+    status = models.CharField(
+        max_length=6, default=NOT_BUILT.colour, choices=ALL_STATUSES
+    )
```
txels: I'm not yet sure how much data we want to store for each build and how. E.g. test results etc.
Also I'd like a second opinion about whether we actually need builds and jobs or just builds.

alexcouper: I'm not sure what the use-case is for Jobs.

Do you have something in mind?

txels: I was literally just turning this diagram into models https://docs.google.com/drawings/d/1tWhxPzuBWYWlxoo4mueNnkUcSG3HrK6691bS6XbQ_xw/edit and found this point and wanted to trigger a discussion. I can get rid of it and add the job name/URL to the build.

txels: Added a `job_id` that should make sense to the CI provider. Then it's up to the provider implementation to store sthg that will make it able to access the job and trigger a build etc.

ployst/builds/models.py
```diff
@@ -0,0 +1,38 @@
...
+    Status = namedtuple('BuildStatus', ['colour', 'result'])
+    Status.__str__ = lambda self: self.result
+
+    SUCCESS = Status('green', 'success')
+    UNSTABLE = Status('yellow', 'unstable')
+    FAILURE = Status('red', 'failure')
+    NOT_BUILT = Status('gray', 'not built')
+    ALL_STATUSES = [NOT_BUILT, FAILURE, UNSTABLE, SUCCESS]
+
+    provider = models.CharField(
```
txels, line 22: Doubting whether we want a model for providers and a foreign key here.
But as providers will need to be added to the codebase, I don't think it really makes sense to manage them in the database.

txels:
I got rid of the Job model, not really needed here. Added data in the build that the provider can use to retrieve job information etc.

ployst/builds/models.py
alexcouper, line 22: Yeah i think this makes sense for now - we can always change in future.

ployst/builds/migrations/0001_initial.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
```
alexcouper, line 1: Currently working on a project that doesn't have the luxury of south.

Glad to have it back!

PR #7, opened 2013-12-14
Dev alex

alexcouper:
@txels This is as far as I got tonight. There are some outstanding questions that I've written in as comments and I''ll highlight them below.


bower.json
```diff
@@ -11,6 +11,14 @@
   "dependencies": {
     "bootstrap": "3.0.2",
     "font-awesome": "4.0.3",
-    "angular": "1.2.1"
+    "angular": "1.2.2",
```
alexcouper, line 14: Not sure why I brought this in - I think you must have done one more commit into master after branching develop?

ployst/core/repos/models.py
```diff
@@ -51,6 +51,15 @@ class Branch(models.Model):
...
     head = Revision(help_text="Latest known revision")
+    # is_contained_by_parent = models.BooleanField(
+    #        help_text="Merged into parent")
+    # parent = models.ForeignKey(Branch, related_name="children")
+    #          - or is this plural and many-to-many?
+    # features = manyToManyWithFeature
+    #            - I think we need this many-to-many relationship so that
+    #              the github app can contain all the logic about whether a
+    #              feature matches a branch. Otherwise at render point we'd
+    #              have to calculate that here again, right?
```
alexcouper, line 63: Question 1 is here. I feel like we're going to need these in order to render the page without having to do further calculations. What do you think?

ployst/github/views.py
```diff
@@ -0,0 +1,61 @@
...
+            - Which feature they relate to?
+        - Update the branch info.
+
+    End goal is that we have enough data at the front end to generate a picture
+    of the following
+
+    - feature branch
+        - dev branch (merged in)
+        - dev branch 2 (not merged in)
+
```
alexcouper, line 39: Ignoring the name of this function for now, is the behaviour documented what we want the `receive_hook` view to have achieved before returning its response?

ployst/github/conf.py
```diff
@@ -0,0 +1,17 @@
+# These two are deliberately out of alphabetical order
+# See the django compressor for an example of using appconf
+# https://github.com/jezdez/django_compressor/blob/develop/compressor/conf.py
+from django.conf import settings  # noqa
+from appconf import AppConf
+
+
```
alexcouper, line 7: Found this as a way to manage app settings.

bower.json
txels, line 14: I think at some point I didn't have network connection to install dependencies and stuck to using versions I already had locally.

ployst/core/repos/models.py
txels, line 63: Seems right. My initial models were not complete, as I didn't have a `features` app. Now I think I accidentally added some of that which I had as "work in progress" so it may be incomplete, no migrations...

This matches our data model anyway: https://docs.google.com/drawings/d/1tWhxPzuBWYWlxoo4mueNnkUcSG3HrK6691bS6XbQ_xw/edit?pli=1 (it would be good to keep the diagrams up to date if we introduce changes, and maybe convert to a format we can put under version control?)

ployst/github/views.py
txels, line 39: Sounds quite right.

Makes me think of a related issue though, which is how to manage the "branching rules" for a repo (branch name conventions, regexes...) if repo set up is per user. So these options should be set at the repo level per team, and one repo can only be set up for a team. Or maybe for now we can have in ployst core multiple instances that ultimately point to the same github repo (in that case a same github repo will have different Repository instances in ployst, where each will can have different settings for different users and a separate copy of a collection of branches - not a big problem, but this process will need to iterate through all those).

txels, line 39: Additionally if we are talking here about "belong to users" then we need to start creating relationships between objects and users in core ployst.

requirements/dev.txt
```diff
@@ -8,3 +9,7 @@ nodeenv
 # documentation
 Sphinx==1.2b3
 sphinx_bootstrap_theme==0.3.2
+
+# Pdb made awesome
+pdbpp
```
txels, line 14: :)

requirements/runtime.txt
```diff
@@ -1,3 +1,8 @@
 django==1.6
 djangorestframework==2.3.5
 South==0.7.6
+
+django-appconf              # provides graceful configuration of apps
```
txels, line 5: Not relevant yet, but for runtime requirements one always pins specific versions.

ployst/github/views.py
alexcouper, line 39: Relationships between users and objects: Yeah we'll need that to be able to write this functionality.

Config: I think we can't realistically enforce a repo being tracked only once site-wide, so it makes sense to me to have the config as you suggest on the repo - with multiple objects in core, all pointing to the same physical repo on disk.

ployst/github/test/test_receive_hook.py
```diff
@@ -0,0 +1,86 @@
...
+from django.test.utils import override_settings
+
+from mock import patch
+
+from . import read_data
+from .. import views  # noqa
+
+
+@override_settings(GITHUB_HOOK_TOKEN='some_valid_token')
+class TestReceiveHook(TestCase):
```
txels, line 14: nice to see the first ployst tests :)
I need to write some for the core APIs.

ployst/core/repos/models.py
```diff
@@ -51,6 +51,15 @@ class Branch(models.Model):
     repo = models.ForeignKey(Repository, related_name='branches')
     name = models.CharField(max_length=100)
     head = Revision(help_text="Latest known revision")
+    # is_contained_by_parent = models.BooleanField(
+    #        help_text="Merged into parent")
+    # parent = models.ForeignKey(Branch, related_name="children")
+    #          - or is this plural and many-to-many?
```
txels, line 57: Some time ago I was giving some thought to representing trees in a relational DB that could be queried in an efficient way and even made a prototype implementation. I may like to give it a try here, if I manage to find the code. Else I can sketch out the idea again.

ployst/core/repos/models.py
```diff
@@ -28,11 +28,11 @@ class Repository(models.Model):
 
     If required in the future, this can be extended to support other
     repo types, such as mercurial, subversion etc.
-
     """
     name = models.CharField(max_length=100)
     url = models.URLField()
     active = models.BooleanField(default=True)
+    local_path = models.CharField(max_length=100)
```
txels, line 35: Reminder: this will need a migration.
While we are in this preliminary phase, I am happy with us updating the initial migration. Although one can argue, with our approach based on very small apps and models that may end up being quite stable (as extension points may end up being JSON fields), migrations will never grow to a large number

PR #8, opened 2013-12-15
Complete project settings for static files

txels:
Hmm I thought I had already included these settings, but apparently not


alexcouper:
i'm sure i merged this the other day too :/

PR #9, opened 2013-12-15
Add initial models for teams, tweak projects

txels:
This adds an initial `accounts` app where we will manage users, teams and team membership. It should match our previous conceptual diagram, but it's open for review and discussion.

For the actual basic user sign up and management functionality (e.g. registration), I am planning to try out http://django-user-accounts.readthedocs.org/

For permissions and object ownership, I'll probably set up some special manager, following a similar scheme to what we use currently at work.

### DIFF
--- a/ployst/core/accounts/models.py
+++ b/ployst/core/accounts/models.py
@@ -12,6 +12,9 @@ class Team(models.Model):
     name = models.CharField(max_length=200)
     users = models.ManyToManyField(User, through='TeamUser')
 
+    def __unicode__(self):
+        return self.name
+
 
 class TeamUser(models.Model):
     """
@@ -35,3 +38,10 @@ class TeamUser(models.Model):
     user = models.ForeignKey(User)
     team = models.ForeignKey(Team)
     role = models.CharField(max_length=1, choices=ROLES)
+
+    def __unicode__(self):
+        return "{user} at {team} ({role})".format (
+            user=self.user.username,
+            team=self.team.name,
+            role=self.get_role_display(),
+        )
--- a/ployst/core/accounts/models.py
+++ b/ployst/core/accounts/models.py
@@ -1,0 +1,37 @@
+from django.db import models
+from django.contrib.auth.models import User
+
+
+class Team(models.Model):
+    """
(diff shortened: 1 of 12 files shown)


ployst/core/accounts/models.py
```diff
@@ -0,0 +1,47 @@
...
+     * Administrator can manage team members
+     * Collaborator can manage team assets, such as projects
+     * Viewer has only read access
+    """
+
+    ADMIN, COLLABORATOR, VIEWER = 'a', 'c', 'v'
+    ROLES = (
+        (ADMIN, 'Administrator'),
+        (COLLABORATOR, 'Collaborator'),
+        (VIEWER, 'Viewer'),
```
txels: thoughts about whether user roles should be per team, per project or per repo?

alexcouper: From skype chat:

Within team, 2 roles:
- admin (to add users)
- normal

Within project, 2 roles:
- admin (set up repos, change provider config etc)
- normal

txels: In that case, what about a simpler structure. Instead of role let's just have:

``` python
manager = models.BooleanField(default=False)
```

And for projects, as users will already be part of a project via the `Team`, we can assume the team admin will be project admin, and we can have additional project admins:

``` python
class ProjectManager(models.Model):
    """
    Users that can manage a project (besides the team managers).

    These can manage project provider settings, but not user permissions,
    which remain the realm of the team managers.
    """
    user = models.ForeignKey(User)
    project = models.ForeignKey(Project)
```

(I just renamed admin to manager as I was just realising that model names started to be the same as classes in `admin.py`.

alexcouper: By team manager in the doc string do you mean team admin?

Sounds good to me

txels: yes I am just calling it manager now as `TeamAdmin` and `ProjectAdmin` exist in admin.py and I found the names confusing. But I can go back to calling it team admin if you prefer.

alexcouper: No preference - continue with manager. just wanted to make sure I
understood correctly (as for a second I thought there might be another
layer of auth)

PR #10, opened 2013-12-22
Writing the github provider pt I

alexcouper:
This replaces PR #7 

Whilst this is still a work in progress, I think it'd be beneficial to merge - providing there aren't any issues with what has been done so far.

Note that I've changed the description of what `recalculate` does (see `github/views.py`).

I also have a question about how we want to handle security of API calls - I'll highlight the question in the PR.


ployst/core/repos/test/test_filtering.py
```diff
@@ -0,0 +1,27 @@
...
+        client = Client()
+        url='http://github.com/pretenders/ployst'
+        response = client.get('/core/repos/repo/?url={0}'.format(url))
+        repos = json.loads(response.content)
+        self.assertEquals(len(repos), 1)
+        self.assertEquals(repos[0]['name'], 'PloystTest')
+
+    def test_only_see_repos_for_my_team(self):
+        "Should not be able to see all repos."
+        # How do we want to handle security in the API.
```
alexcouper: Here is my security question...

ployst/github/views.py
```diff
@@ -0,0 +1,62 @@
...
+
+def recalculate(repo, branch_ref):
+    """
+    Recalculate all relevant branch information given a new commit
+
+    Encapsulated in this function:
+
+    The full update given a repo is:
+        - Get the team(s) from the GITHUB_HOOK_TOKEN used.
+        - Get all features that are active and belong to these teams.
```
alexcouper, line 23: 22 and 23 are different from the original PR. 

This will mean having the GITHUB_HOOK_TOKEN unique per team - but I think that is something we'll want to do anyway.

alexcouper:
Warning: This PR is starting to get pretty big...

ployst/core/features/test/test_api.py
```diff
@@ -0,0 +1,19 @@
...
+from django.test.client import Client
+
+class TestProjects(TestCase):
+    fixtures = ['teams.json', 'projects.json']
+
+    def test_get_projects_by_team_guid(self):
+        "Test we can get a team by id"
+        client = Client()
+        response = client.get(
+            '/core/features/project/?team=8d7de2c4-4849-452f-82af-e142641c4b6d'
```
alexcouper, line 14: After this, I'll get the features for each project for use in `github/views.py`.

@txels is there anyway of definining a filter such that it works across relationships - that way I could do something like:

```
/core/features/feature/?team=team-id
```

and get everything back in a single request.

alexcouper, line 14: **Update**

I'm now writing the api call to get the list of features for a project.
Ideally I'd like to be able to do this:

```
/core/features/project/{id}/features
```

rather than following my earlier approaches and doing this:

```
/core/features/feature/?project_id={id}
```

Not sure if it's possible, but going to look into it..

alexcouper, line 14: Ah looks like to do that I need to use the [drf-nested-routers package](https://github.com/alanjds/drf-nested-routers/).

It's referred to in the docs and was deliberately kept out of the core django-rest-framework [here](https://github.com/tomchristie/django-rest-framework/pull/1048).

Do you have a view on bringing in this dependency?

alexcouper, line 14: I made an attempt at using this in a separate branch, and it turns out that it's by no means a silver bullet.

For now, I have continued with the explicit filtering approach.

I think at this point we should think about how we want to lock the API down so that, for example, user a in team 101 can't GET at the projects that team 102 has.

ployst/core/client/accounts.py
```diff
@@ -0,0 +1,11 @@
+import requests
```
alexcouper, line 1: I think for now it might be better to have all of these client methods in a single client.py module, probably within a `Client` class.

ployst/core/accounts/fixtures/teams.json
```diff
@@ -0,0 +1,16 @@
+[
```
txels, line 1: can we put test fixtures somewhere under some `test` directory?

ployst/core/accounts/models.py
```diff
@@ -9,12 +11,18 @@ class Team(models.Model):
     Projects are assigned to teams. Users may belong to many teams.
 
     """
+    guid = models.CharField(max_length=50, primary_key=True, editable=False)
```
txels, line 14: can you explain the rationale behind this in specifically this model?

ployst/core/client/accounts.py
txels, line 1: yes it sounds appropriate.

ployst/core/accounts/models.py
txels, line 14: (if we want to generalise this it would be useful to go for a custom field such as https://djangosnippets.org/snippets/1262/)

alexcouper, line 14: Sure - I wanted a field that could be team specific and be used when connecting to services such as github as a token identifier.

It may be that in time we want to move this out of Team into a separate model of "tokens" or something with a team having many, but I thought for now it would be ok with a team having only one such token.

Once I had settled on only having one, it made some sort of sense to make this the primary key. But that's by no means a sticking point.

ployst/core/client/accounts.py
txels, line 1: BTW did you ever get to see this? http://technomilk.wordpress.com/2013/03/25/apitopy-a-pythonic-way-to-access-http-apis/ My pythonic _universal_ API client - do you think it could be of use here? I'd love to extend that with a pythonic _write_ syntax as well.

ployst/core/features/test/test_api.py
txels, line 14: Yes that is what I see as my next priority task, object ownership/permissions.

txels, line 14: BTW I would not make filtering by "team" (or ownership) part of the API signature, I think it should happen as an implicit part of the filtering based on authenticated user that should happen on every request.

ployst/core/repos/migrations/0001_initial.py
```diff
@@ -22,7 +22,7 @@ def forwards(self, orm):
             (u'id', self.gf('django.db.models.fields.AutoField')(primary_key=True)),
             ('repo', self.gf('django.db.models.fields.related.ForeignKey')(to=orm['repos.Repository'])),
             ('name', self.gf('django.db.models.fields.CharField')(max_length=100)),
-            ('head', self.gf('ployst.repos.models.Revision')(max_length=40)),
+            ('head', self.gf('ployst.core.repos.models.Revision')(max_length=40)),
```
txels, line 25: ah yes I realised about this one as well.

ployst/core/repos/models.py
```diff
@@ -28,11 +29,11 @@ class Repository(models.Model):
 
     If required in the future, this can be extended to support other
     repo types, such as mercurial, subversion etc.
-
     """
     name = models.CharField(max_length=100)
     url = models.URLField()
     active = models.BooleanField(default=True)
+    local_path = models.CharField(max_length=100)
```
txels, line 36: Should we make this `null=True`? This will only exist after the repo has been cloned, but I'm thinking we may create instances of this model before triggering a cloning.

OTOH, `local_path` could be a property and a location of a repo working copy on local storage may be computed based on a convention and some setting. Also, we may want to think that file storage may not even be local (if we eventually want to make ployst load-balanceable).

ployst/core/repos/models.py
```diff
@@ -51,6 +52,10 @@ class Branch(models.Model):
     repo = models.ForeignKey(Repository, related_name='branches')
     name = models.CharField(max_length=100)
     head = Revision(help_text="Latest known revision")
+    is_contained_by_parent = models.BooleanField(
+           help_text="Merged into parent")
```
txels, line 56: why not name the field `merged_into_parent` then?

ployst/core/features/test/test_api.py
alexcouper, line 14: I agree - so a Github provider has credentials and has access across all teams.

ployst/core/repos/models.py
```diff
@@ -51,6 +52,10 @@ class Branch(models.Model):
...
+    parent = models.ForeignKey("self", related_name="children", null=True)
```
txels, line 57: I'm giving a thought to not using self references to implement hierarchies. I had some experimental code for that but can't find it anywhere.

ployst/core/features/test/test_api.py
alexcouper, line 14: Oh I see you may be saying something different.

The filtering above needs to happen because the github provider only wants to look at projects owned by a particular team. Or do you forsee the github provider authenticating "as" that team before requesting for projects?

ployst/core/repos/test/test_api.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def test_only_see_repos_for_my_team(self):
+        "Should not be able to see all repos."
+        # How do we want to handle security in the API.
+        # Should all requests be sent with some kind of team token?
+        # Should the backend (ie providers) also need to use such a token?
+        raise NotImplementedError("test_only_see_repos_for_my_team")
+
+
+
```
txels, line 27: blank lines... pep8... ;)

ployst/core/repos/test/test_api.py
```diff
@@ -0,0 +1,27 @@
...
+        url='http://github.com/pretenders/ployst'
+        response = client.get('/core/repos/repo/?url={0}'.format(url))
+        repos = json.loads(response.content)
+        self.assertEquals(len(repos), 1)
+        self.assertEquals(repos[0]['name'], 'PloystTest')
+
+    def test_only_see_repos_for_my_team(self):
+        "Should not be able to see all repos."
+        # How do we want to handle security in the API.
+        # Should all requests be sent with some kind of team token?
```
txels, line 22: I think we want all requests authenticated, and an implicit mechanism will only retrieve objects I am allowed to access.

ployst/github/conf.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class GithubConf(AppConf):
+    """
+    Config settings for this app.
+
+    .. note::
+
+        ``AppConf`` will auto convert these to be eg. <appname>_HOOK_TOKEN for
+        use in settings. In this case GITHUB_HOOK_TOKEN.
```
txels, line 15: unrelated Q, but is the hook token a global setting or something user-account-related?

alexcouper, line 15: It was global originally, but has now become user account related. Or in actual fact guid of team. I suspect we want to abstract it out into some additional model that has a relationship with team.

ployst/github/views.py
txels, line 23: Yes it makes sense to use per team tokens since the beginning.

txels:
Done a quick first read with the little time I had. Will try to be more thorough and deep sometime later, and give it a play.

ployst/github/conf.py
txels, line 15: Yes within some model that contains team settings.
Actually it could just be a JSON field in the `Team` model. But we can make it cleaner by having `TeamSettings` one instance per team and provider, so that providers do not conflict.

ployst/core/features/test/test_api.py
txels, line 14: Oh I see. I'll think about it on a second thorough review.

ployst/core/repos/migrations/0002_auto__add_field_repository_local_path__add_field_branch_is_contained_b.py
```diff
@@ -0,0 +1,138 @@
...
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        # Adding field 'Repository.local_path'
+        db.add_column(u'repos_repository', 'local_path',
+                      self.gf('django.db.models.fields.CharField')(default='MISSING', max_length=100),
```
txels, line 13: If this field is not to become `null=True`, a default of `''` would also work and is an easier to handle default.

requirements/runtime.txt
```diff
@@ -1,6 +1,12 @@
...
 django-compressor==1.3
+requests==2.1.0
 # slimit==0.8.1               # used by django compressor for compressing js files
 
 South==0.7.6
+
+django-appconf==0.6         # provides graceful configuration of apps
+django-filter==0.7          # provides generic filtering in API
+# Git provider requirements
+GitPython==0.1.7            # access git repositories
```
txels, line 12: can we sort requirements in some sensible way before they grow large? Alphabetically might work, but also putting together 3rd party django apps and 3rd party python libs, similarly as we tend to to for imports.

requirements/dev.txt
```diff
@@ -1,4 +1,5 @@
 -r runtime.txt
+-r test.txt
```
txels, line 2: hm the approach how I intended the requirements is that one file only is required per environment, and includes all that is needed.
That means I would move the first line `-r runtime.txt` from here to the beginning of `test.txt`.

txels:
Although there are some open comments, I am happy to merge this in the current state. Although there is one failing test due to a fixture that was not uploaded to version control. I am working on object ownership in a separate branch, I will decide between keeping this one open or merging it depending on whether my branch will benefit from this being already in.

txels:
I decided to merge and build from here.

ployst/core/features/test/test_api.py
alexcouper, line 14: Did you get any further with your thoughts on this?

txels, line 14: The code in my permissions PR includes now a manager that will filter data per team. We can use that in the API but we will have to customise the views' queryset. This as an example is the kind of mixin we use at Rockabox for this type of filtering (our viewsets extend this):

``` python
class PermissionsMixin(object):
    def get_queryset(self):
        objects = super(PermissionsMixin, self).get_queryset()
        if hasattr(objects, 'filter_for_user'):
            return objects.filter_for_user(self.request.user)
        else:
            return objects.all()
```

In the GH hook case, we can adapt the queryset based on the querystring and use the `for_team` method?
In other views where we filter by current user we can apply to the logged in user.

Does this make sense?

requirements/runtime.txt
alexcouper, line 12: sorted.

ployst/core/repos/migrations/0002_auto__add_field_repository_local_path__add_field_branch_is_contained_b.py
alexcouper, line 13: Changed

ployst/core/repos/test/test_api.py
alexcouper, line 27: Dealt with

ployst/core/repos/models.py
alexcouper, line 57: Still waiting on this I guess...

ployst/core/repos/models.py
alexcouper, line 56: Changed, and updated the migrations along with it...

ployst/core/accounts/fixtures/teams.json
alexcouper, line 1: irrelevant now

ployst/core/features/test/test_api.py
alexcouper, line 14: So a question is the following.

The github PR hook will receive a request from github from which we will derive which team's repo was updated.

We then need to query core's API to get feature information back for that team. We won't have a session at this point, or if we do it will be a more generic session. What request should be sent to core to get the set of features applicable for a team? I presume we don't want to force the github PR hook to emulate creation of a session as the user.

ployst/core/accounts/models.py
txels, line 14: yeah I like the idea of this being a separate "token" one could eventually revoke.

ployst/core/features/test/test_api.py
txels, line 14: No you're right. The providers should have privileged access to the core API without session authentication.

If we think of Tokens as a security access code, then they are an alternative secured access to session authentication, so you really do not need both.

My idea was implement some custom authentication in DRF similar to http://www.django-rest-framework.org/api-guide/authentication#tokenauthentication but URL-based.

The slight mismatch is that we do not think of a token as an alternative way to identify a User, but a Team. Maybe that is not what we really want, and a user-associated token might be simpler to manage, as it maps more directly to built-in authentication?

Also, as this is going to be a token in a single request, there is no need for a "session" (something that spans multiple requests). Maybe we need to implement custom permissions that go together with authentication: http://www.django-rest-framework.org/api-guide/permissions

ployst/core/repos/models.py
txels, line 57: :]

so let's go with parent for now...

ployst/core/features/test/test_api.py
alexcouper, line 14: I'm going to raise a separate issue for this so that we don't lose this.

alexcouper, line 14: Added as issue #15

ployst/github/conf.py
alexcouper, line 15: Added as issue #16

ployst/core/accounts/models.py
alexcouper, line 14: Incorporated (possibly wrongly) in issue #16

requirements/dev.txt
alexcouper, line 2: dealt with

PR #11, opened 2013-12-30
Add basis for object ownership checks

txels:
I've added a base manager for models that represent data owned by teams. These models must have a simple field `team_lookup` that is a django ORM lookup to the team that owns that object. With that in place, the manager adds a method `for_team(team)` that returns a queryset that can be used as the base for further lookups. This queryset filters out all data that does not belong to the given team.

Additionally, I started using factories for test objects instead of fixtures.


ployst/core/repos/models.py
```diff
@@ -30,11 +31,14 @@ class Repository(models.Model):
     If required in the future, this can be extended to support other
     repo types, such as mercurial, subversion etc.
     """
+    project = models.ForeignKey(Project, related_name='repositories')
```
txels, line 34: I am not sure this is the right relationship.

We may want to separate `Project` (as an entity in ployst to group stuff together coming from various providers) from `PlanningProject` as project within the planning tool of choice. Project will then be an entity created and defined within ployst by a team manager to group repos, planning projects, etc. for a team.
Thoughts?

alexcouper, line 34: I'm not clear what purpose `PlanningProject` serves to be honest.

But certainly the only `Project` that a `Repository` has a relationship with is that which groups features and repos together. 

Arguably teams (plural) should be includable within a ployst project - as opposed to the team being within a project.

txels, line 34: `PlanningProject` (if it ever happens) is the container where features stemmed from in the planning tool. E.g. with TP at Rockabox we have a number of projects, each of them with its user stories, bugs, etc.

It may be that we do not need to track this and it's just a configuration thing for project and provider.

I was wondering at this point because the `Project` model we have now is part of the features app, which seemed to be the domain of the planning provider, so I was thinking of that one as what exists in the planning tool. Maybe `Project` does not belong in `features` but in `accounts`.

Re: teams and projects - now projects are assigned to teams in a 1 team - many projects, but maybe this should be a many to many.

txels:
There are a number of tests failing because of fixture-related problems.
I will fix those by moving test data creation to factories instead of fixtures, if you're OK with that.

alexcouper:
Fine by me

txels:
Now all pre-existing passing tests are migrated to use factories, and pass.

ployst/core/repos/models.py
alexcouper, line 34: Yeah I think `Project` belonging in `accounts` makes sense.

txels, line 34: OK will move now.

PR #14, opened 2014-01-20
Move project to accounts app

txels:
This looks like a lot of changes, but it's only the result of moving `Project` to the accounts app.

I for now squashed all migrations into one per app. At some point we can decide that the models are stable enough, or we have enough local test data that we want to start properly migrating stuff. Until then we can use `schemamigration --auto --update`.


ployst/core/repos/test/test_api.py
```diff
@@ -1,21 +1,24 @@
...
         Test we search by url to get a list of repos.
 
         """
-        url = 'http://github.com/pretenders/ployst'
-        repo = RepositoryFactory(name='PloystTest', url=url)
+        repo_url = 'http://github.com/pretenders/ployst'
+        repo = RepositoryFactory(name='PloystTest', url=repo_url)
 
-        response = self.client.get('/core/repos/repo/?url={0}'.format(url))
+        # url = reverse('repos:repository-list')
```
txels: For some reason this reverse doesn't work. I can't see what I'm doing wrong.

txels: Oh it was doubly namespaced. Fixed!

alexcouper:
Squashing all migrations makes a lot of sense to me.

PR #17, opened 2014-01-24
Move client methods into simple module from package.

alexcouper:


alexcouper:
All tests pass ('cept the `NotImplementedError` one)

PR #20, opened 2014-02-06
Alex/280 github provider

alexcouper:
This almost completes the github provider flow. At the very least it shows the way it is going to happen.

The test found in test_end_to_end.py, which mocks out core responses shows it "working". However there are a few missing pieces before it can be used in reality:
- [x] Provider-specific settings. This concept is missing, as is an api call.
- [x] Updating branch information. This API call is missing.

Once these are done, the github provider should be working.

However, until we have the following, it won't be particularly useful:
- Ability to select the repos to monitor
- A public facing address to which github can POST

Additionally, I think we should do the following clean up after we have got this working fully:
- [ ] Make the recalculate method a background running task to be taken care of after responding to github.
- [x] Tidy up the client to be a class with a url (and token) passed in to the constructor.


ployst/core/client.py
```diff
@@ -25,10 +15,20 @@ def get_features_by_project(project_id):
 
 
 def get_repos_by_url(url):
-    # I can't seem to get reverse to work for django rest framework.
-    # Any hints?
```
txels, line 29: Ha yes URL pattern names for DRF model viewsets are not the best documented of things.
See here for an example: https://github.com/pretenders/ployst/blob/develop/ployst/core/features/test/test_api.py#L24 

Patterns are `<model>-list` and `<model>-detail`

txels:
Looks like a nice substantial piece of work. I'll properly review this within the next days.

ployst/settings/dev.py
```diff
@@ -3,3 +3,7 @@
 
 DEBUG = True
 TEMPLATE_DEBUG = True
+TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
+INSTALLED_APPS += (
+    'django_nose',
+)
```
txels: Oh you've reintroduced the nose test runner? It was left out on purpose in favour of the newer unittest test runner in django 1.6 that has basically the same test discovery mechanism. What is the limitation you found?

https://docs.djangoproject.com/en/1.6/releases/1.6/#discovery-of-tests-in-any-test-module

We still can use `nose.tools` without the nose test runner if that's what you're after.

ployst/core/client.py
```diff
@@ -25,10 +15,20 @@ def get_features_by_project(project_id):
...
     )
     return response.json()
 
+
+def get_provider_settings(team_id, provider):
+    #TODO: make this API call.
+    if provider == "github":
+        return {
+            "branch_finders": ["^master$", ".*(?i){feature_id}.*"]
+        }
```
txels, line 53: Added a story for supporting provider settings per team: http://txels.tpondemand.com/entity/319
We need to add models and endpoints (viewsets) for that.

ployst/core/client.py
```diff
@@ -25,10 +15,20 @@ def get_features_by_project(project_id):
...
+
+
+def update_branch_information(branch_info):
+    # TODO: write this API call.
+    pass
```
txels: Should be doable with current endpoints?

ployst/core/repos/models.py
```diff
@@ -46,6 +46,11 @@ class Meta:
     def __unicode__(self):
         return self.name
 
+    @property
+    def team(self):
```
txels, line 50: This property could be easily added to base class `TeamObject` in a generic way, using the already defined `team_lookup`, and it would be available for all _owned_ objects.

ployst/github/git_adapter.py
```diff
@@ -0,0 +1,60 @@
+import git
+
+
+class GitAdapter(object):
```
txels, line 4: The purpose of this seems to be to have multiple implementations. Can you explain the rationale for it?

ployst/github/models.py
```diff
@@ -0,0 +1,96 @@
+class Hierarchy(list):
```
txels: `models.py` makes one erroneously think these will be django models (in places where these are imported).
I find it a bit confusing, but after that initial mental adjustment it's fine I guess.

ployst/github/models.py
```diff
@@ -0,0 +1,96 @@
...
+        level = hierarchy_level.level
+        node.parent = self.get_parent(level)
+
+        if len(hierarchy_level) == 1:
+            for child_node in self[level + 1]:
+                child_node.parent = node
+            self.insert(level, hierarchy_level)
+
+
+class HierarchyLevel(list):
```
txels: a little docstring might help

ployst/github/models.py
```diff
@@ -0,0 +1,96 @@
...
+        self.commit = commit
+        self._parent = None
+        self.children = []
+        self.in_parent = None
+
+    @property
+    def parent(self):
+        return self._parent
+
+    @parent.setter
```
txels: nice reminder that one can use this to create a setter for a property :)

ployst/github/models.py
```diff
@@ -0,0 +1,96 @@
...
+    def parent(self, value):
+        if self._parent:
+            self._parent.children.remove(self)
+        self._parent = value
+        if self._parent:
+            self._parent.children.append(self)
+
+    def __eq__(self, other):
+        # import pdb; pdb.set_trace()
```
txels: :)

ployst/github/models.py
```diff
@@ -0,0 +1,96 @@
...
+        if type(other) == tuple:
```
txels: what about `isinstance`?

ployst/github/models.py
```diff
@@ -0,0 +1,96 @@
...
+            return (self.branch, self.commit) == other
+        elif type(other) == HierarchyNode:
```
txels: same here.

ployst/github/test/test_end_to_end.py
```diff
@@ -0,0 +1,87 @@
+from django.core.urlresolvers import reverse
+from django.test import TestCase
+from mock import patch, Mock
+
+from . import read_data, DUMMY_REPO, ensure_dummy_clone_available
+from .. import views  # noqa
+
+
+class MockClient(object):
```
txels, line 10: nice

ployst/settings/dev.py
alexcouper: Actually I was after the `-s` option that nose has (don't capture output), but I've just tried it without the nose runner and turns out I can still get the kind of output I want (pdb). This will be removed in my next commit...

ployst/github/git_adapter.py
alexcouper, line 4: The purpose was to define an interface to be used when interfacing with Git. 

The reason for this was that I was (and am) unsure which approach is better: using the python implementation of git in PyGit or using git from the command line.

Perhaps this was done too soon, given that there is currently only one implementation of the `GitAdapter`.

Thoughts?

ployst/github/models.py
alexcouper: I was struggling for module names. Suggestions more than welcome!

ployst/github/models.py
alexcouper: Definitely better!

ployst/apibase/permissions.py
```diff
@@ -13,7 +14,8 @@ def contains_valid_token(request):
     `X-Ployst-Access-Token` to match a token for a registered client.
 
     """
-    access_token = request.META.get(TOKEN_HEADER)
+    access_token = request.META.get(_HTTP_TOKEN_LOOKUP)
```
txels: :) good point

ployst/apibase/permissions.py
```diff
@@ -4,6 +4,8 @@
 
 TOKEN_HEADER = 'X-Ployst-Access-Token'
 
+_HTTP_TOKEN_LOOKUP = "HTTP_" + TOKEN_HEADER.replace('-', '_').upper()
```
txels: not sure why the leading `_`

ployst/core/client.py
```diff
@@ -1,34 +1,61 @@
...
+class Client(object):
+
+    def __init__(self, base_url, access_token):
+        self.base_url = base_url
+        self.access_token = access_token
+
+    def get(self, path):
+        response = requests.get(
+            '{0}/{1}'.format(self.base_url, path),
+            headers={'X-Ployst-Access-Token': self.access_token}
```
txels, line 16: This is no longer the right header, should be `_HTTP_TOKEN_LOOKUP`.
BTW I've now implemented headers in https://github.com/txels/apitopy as per your request, and also initial support for additional HTTP verbs. Tested successfully against this branch with `GET`, `POST` and `DELETE`.

ployst/github/models.py
alexcouper: Hopefully this is clear now at the top of the module...

ployst/apibase/permissions.py
alexcouper: It was originally because I expected this to be something that wasn't used elsewhere. Then tests happened.

Changing...

ployst/core/client.py
alexcouper, line 16: I think it should be `TOKEN_HEADER` not `HTTP_TOKEN_LOOKUP` because I believe it is django that converts it to have the HTTP_ preceding the header name, no?

I'll be looking into the apitopy stuff now.

txels, line 16: When I tested this I actually ended up using:

``` python
api = apitopy.Api(
    'http://localhost:8000/core/', verbose=True, ensure_slash=True,
    headers=headers)
headers = {'X_PLOYST_ACCESS_TOKEN': '27665cd2-c386-4fa6-a741-656fcf23ca65'}
api.repos.repo()
```

The `ensure_slash` I had to add to make sure apitopy used a trailing slash for URLs when posting to them. Else it wouldn't work, as Django would want to redirect me to a URL with trailing slash, but complained that it couldn't work with a POST.

ployst/core/client.py
```diff
@@ -1,34 +1,56 @@
...
+
+    def get_provider_settings(self, team_id, provider):
+        #TODO: make this API call.
+        if provider == "github":
+            return {
+                "branch_finders": ["^master$", ".*(?i){feature_id}.*"]
+            }
+
+    def update_branch_information(self, branch_info):
+        self.post('core/repos/branch/', branch_info)
```
txels: it should now be possible to do:

``` python
self.ployst.core.repos.branch.POST(data=branch_info)
```

...but you will have to initialise `Api` with `ensure_slash=True` (dirty trick).

ployst/core/client.py
```diff
@@ -1,34 +1,56 @@
...
-        'http://localhost:8000/core/accounts/team/{0}'.format(guid)
-    )
-    if response.status == 200:
-        return True
+class Client(object):
 
-    return False
+    def __init__(self, base_url, access_token):
+        assert base_url.endswith('/')
+        self.base_url = base_url
```
txels: if you set the base_url to include `/core/` at the end it saves you from having to include `.core` on every client request ;)  (e.g. `self.ployst.repos.repo[3].DELETE()`) or `self.ployst.features.project(team=team_id)`

As this is the client for `core` we do not need to access other URLs outside that path.

ployst/core/client.py
```diff
@@ -1,34 +1,56 @@
...
+        )
+        return response.json()
 
-def get_projects_by_team(team_id):
-    response = requests.get(
-        'http://localhost:8000/core/features/project/?team={0}'.format(team_id)
-    )
-    return response.json()
+    def get_features_by_id(self, feature_id):
+        return self.ployst.core.features.feature(feature_id=feature_id)
```
txels: I can't help to like how this looks ;)

OTOH we have the downside of not using `reverse` for URLs --- as an upside, we can package the client separately from core, if we do not import from it.

ployst/core/client.py
txels: In the end I'd love to be able to do:

``` python
self.ployst.core.repos.branch += branch_info  # POST
self.ployst.core.repos.repo[2] = repo_info  # PUT or maybe PATCH
del self.ployst.core.features.feature[1234]  # DELETE
```

(this is kind of my dream API for `apitopy`, that every HTTP operation looks like a pythonic construct)

ployst/core/features/views.py
```diff
@@ -9,4 +9,4 @@
 class FeatureViewSet(PermissionsViewSetMixin, ModelViewSet):
     model = Feature
     serializer_class = FeatureSerializer
-    filter_fields = ('project',)
+    filter_fields = ('project', 'feature_id')
```
txels, line 12: Why do you need to filter by feature ID? It is included in the viewset, that you can do `/features/feature/<id>`.

ployst/core/repos/serializers.py
```diff
@@ -13,7 +13,9 @@ class Meta:
 class RepositorySerializer(DynamicFieldsSerializerMixin,
                            serializers.ModelSerializer):
     branches = BranchSerializer(many=True)
+    team = serializers.RelatedField(many=False)
```
txels, line 16: I had to add `read_only=True` to both of these, in order to be able to POST a repo.

ployst/core/repos/serializers.py
```diff
@@ -13,7 +13,9 @@ class Meta:
...
 
     class Meta:
         model = Repository
-        fields = ('id', 'name', 'branches', 'url')
+        fields = ('id', 'name', 'branches', 'url', 'project',
+                  'team', 'local_path')
```
txels, line 21: I'm unsure about `local_path` and its use. As this path will be local to the entity doing the cloning (the github provider in this case), do we really need to store it in core? Can we not build a path based on some convention, so that repo name or id is enough for the provider to manage its clones?

ployst/github/git_adapter.py
txels, line 4: I understand now.

So do you think we will end up with multiple implementations living together, or will one replace the other?
Or asked differently, do we want to support both and let people decide whether they want to install `PyGit` optionally and fall back to command line if not found?

ployst/github/git_adapter.py
```diff
@@ -0,0 +1,60 @@
...
+        branches = []
+        for remote_ref in self.remote.refs:
+            if remote_ref.remote_head == "HEAD":
+                continue
+            branches.append((remote_ref.remote_head,
+                            str(remote_ref.commit)))
+        return branches
+
+    def is_contained(self, commit, branch):
+        return commit in [c.hexsha for c in self.repo.iter_commits(branch)]
```
txels: this is probably more readable than the git command line alternative.

txels: more efficient if you use a generator `return commit in (c.hexsha for c in self.repo.iter_commits(branch))`

ployst/github/lib.py
```diff
@@ -0,0 +1,80 @@
...
+        return statuses
+
+
+def match_features(features, regexes, branch_name):
+    """
+    Get a list of features that match the branch_name using the regex given
+    """
+    matched = []
+    for feature in features:
+        feature_id = feature['feature_id']
```
txels, line 76: if `features` has been fetched using `apitopy` then you can access it with `feature.feature_id` (thanks to `dotify`).

ployst/github/models.py
txels: `trees` or `hierarchy` work for me.

In the case of modules that don't have to follow django conventions, I'd prefer the module name to express the subset of functionality included rather than some sort of classification of type of code (e.g. try to avoid generic names like `mixins` or `bases` or `classes` or `lib` - I am guilty of that approach as well sometimes, in particular `mixins`). It goes a bit against what django people tend to do (as django conventions put together things by _type_) but I find it better to group related functionality together - it gives better isolation between modules.

Open to opinions.

ployst/github/models.py
txels: Yes that's quite good.

ployst/core/client.py
alexcouper, line 53: I've done this as part of this PR now...

ployst/github/models.py
alexcouper: Happy with this approach, and have renamed models to be trees.py

ployst/core/client.py
alexcouper: That would be very cool.

For now, I've only implemented the GET requests using the apitopy as I need both PUT and POST so I'm leaving those in requests land for now...

ployst/core/client.py
alexcouper: Seems reasonable. Changing...

ployst/github/test/__init__.py
```diff
@@ -1,7 +1,28 @@
...
+
+
+def ensure_dummy_clone_available():
+    """
+    Check that we have access to pretenders' dummyrepo
+    """
+    if not exists(DUMMY_CODE_DIR):
+        os.mkdir(DUMMY_CODE_DIR)
+    folder_name = join(DUMMY_CODE_DIR, 'dummyrepo')
+    if not exists(folder_name):
```
txels: This fails out of the box because the `dummyrepo` dir exists (it's under version control here) but it doesn't have a `.git` directory inside.

So since the folder exists, it's not cloned, and the fetch command later fails to find `dummyrepo/.git`.

You probably at some time cloned `dummyrepo` manually and it just worked since.

ployst/github/test/data/dummyrepo
```diff
@@ -0,0 +1 @@
+Subproject commit 7098fa31bf9663343c723d9d155c0dc6e6e28174
```
txels: The culprit for my previous comment.
You started using subprojects and later on changed your mind?

ployst/core/client.py
alexcouper: I eventually decided it was better not using `reverse` for that reason - so this is great!

ployst/core/features/views.py
alexcouper, line 12: `feature_id`, as opposed to `id`.

Where `feature_id` is some string representation of a feature, eg `"us101"` tied to a project - it may not be unique across all projects.

ployst/github/views.py
```diff
@@ -36,30 +40,74 @@ def recalculate(repo, branch_ref):
...
     """
-    # projects = client.features.get_projects_by_team(token)
-    # for proj in projects:
-    #     client.features.get_features_by_project(proj['id'])
+    repos = client.get_repos_by_url(repo_url)
+    for repo in repos:
+
+        #TODO: Implement settings. Model and API calls.
+        prov_settings = client.get_provider_settings(
+            repo['team'],
```
txels: can probably use `repo.team`

ployst/github/git_adapter.py
alexcouper: Altered

ployst/github/views.py
```diff
@@ -36,30 +40,74 @@ def recalculate(repo, branch_ref):
...
         commit_info = json.loads(payload)
         url = commit_info['repository']['url']
-        branch_ref = commit_info['ref']
+        branch_name = commit_info['ref'].replace('refs/heads/', '')
     except (KeyError, ValueError):
         LOGGER.error('Unexpected data structure: {0}'.format(request.POST))
         return HttpResponseBadRequest()
 
-    recalculate(url, branch_ref)
+    # TODO: Make this an async task.
```
txels, line 111: as a separate story... anything in mind?

ployst/github/test/test_git_adapter.py
```diff
@@ -0,0 +1,44 @@
...
+        The dummy repo used has a master branch that is contained in all sub
+        branches, but that does not contain the HEAD commits of any of them.
+        """
+        gh = PythonGitAdapter(DUMMY_REPO)
+
+        # HEAD of master should be in master and all other branches.
+        assert_true(gh.is_contained('0f6eefefc14f362a2c6f804df69aa83bac48c20b',
+                                    'master'))
+        assert_true(gh.is_contained('0f6eefefc14f362a2c6f804df69aa83bac48c20b',
+                                    'my/feature-99'))
```
txels, line 39: This test fails for me:

```
======================================================================
FAIL: test_is_contained (ployst.github.test.test_git_adapter.TestPythonGitAdapter)
----------------------------------------------------------------------
  File "/home/carles/dev/mine/ployst/ployst/github/test/test_git_adapter.py", line 39, in test_is_contained
    'my/feature-99'))
AssertionError: False is not true
```

Some data for diagnosis:

```
$ cd /home/carles/dev/mine/ployst/ployst/github/test/data/dummyrepo
$ git fetch
$ git branch
* master
$ git branch -a
* master
  remotes/origin/HEAD -> origin/master
  remotes/origin/master
  remotes/origin/my/feature-99
  remotes/origin/my/feature_branch
  remotes/origin/story/101/fred
$ git log  # with some options...
* 7098fa3 (HEAD, origin/my/feature-99, my/feature-99) Add an extra line.
* 0f6eefe (origin/master, origin/HEAD, master) Create README.md
```

ployst/github/test/__init__.py
alexcouper: Fixed - and actually this exposed another issue in the tests which I've also fixed.

ployst/github/views.py
alexcouper: This was actually wrong, it should be project.

ployst/github/test/test_git_adapter.py
alexcouper, line 39: This is the one I was referring to above - sorted now.

ployst/github/views.py
alexcouper, line 111: Was thinking celery - but open to alternatives.

alexcouper:
I think I've dealt with all comments.

If not, let me know and I'll come back in to this.

Once this has been merged I'll look at asyncing the github hook process.

txels:
It all looks good. I'm trying to make the tests run, but there is one simple failure I can fix myself and merge

ployst/core/repos/models.py
txels, line 50: Made a PR for this: https://github.com/pretenders/ployst/pull/21

ployst/core/features/views.py
txels, line 12: True. Oops.

PR #21, opened 2014-02-18
Make team property available to all TeamObject subclasses

txels:
This stems from a comment I made to your PR, how it should be easy to make a `team` property available to all subclasses of `TeamObject`.

### DIFF
--- a/ployst/core/accounts/models.py
+++ b/ployst/core/accounts/models.py
@@ -53,13 +53,23 @@ class TeamUser(models.Model):
 
 class TeamObject(models.Model):
     """
-    Base class for models that are objects that are owned by a team
+    Base class for models that are objects that are owned by a team.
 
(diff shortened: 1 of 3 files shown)


ployst/core/accounts/models.py
```diff
@@ -53,14 +53,24 @@ def __unicode__(self):
...
     class Meta:
         abstract = True
 
+    @property
+    def team(self):
+        team_path = self.team_lookup.split('__')
+        value = self
+        for attr in team_path:
+            value = getattr(value, attr)
+        return value.guid
```
txels, line 72: we may want to improve DB efficiency adding some `select_related` to the queryset for `TeamObjectsManager`.

txels, line 72: (added this in a second commit)

alexcouper:
Nice!

PR #22, opened 2014-02-23
UI: Login, Logout some HTML and CSS

txels:
@alexcouper: this is work in progress, so that you know what I'm working on, but enough of it actually works to consider merging.

This introduces user login and logout, as well as the initial structure for the UI:
- Bootstrap3
- compressor
- Some base template for user account management (this page will host an angular app)

There are some things that have been copied over from `deploystream`, I have reviewed some of it to use Bootstrap3.

I chose to use http://django-crispy-forms.readthedocs.org/ to properly lay out forms using bootstrap3.

After running `fab develop` to update dependencies, it should work out of the box: browse to http://localhost:port/ and should show a login page, etc. Let me know if it doesn't.


alexcouper:
It would be nice if the fab command could check if there is a nodeenv available and offer to run `nodeenv -p` if not.

alexcouper:
The selector on the left hand side of the page is broken in the sense that it doesn't highlight `Teams` or `Provider` when they are selected.

alexcouper:
Happy to merge as is though

PR #23, opened 2014-03-02
Kick off async process on github posts. Add command for easier testing.

alexcouper:
This PR contains some controversial elements.

The main two being:
- the decision to make celery a system wide runner and not just nestled inside the github package.
- the addition of github as a django `INSTALLED_APP`

I began by writing a celery app just within the github app, but decided that it made more sense for the github app to define the tasks it needs, but not be responsible for running celery itself.

Additionally, if settings is just seen as the final configuration layer when setting up ployst, and it can contain settings like `GITHUB_HOOK_TOKEN`, then it makes sense for github to be an app listed in `INSTALLED_APPS`. (And this had the benefit of making it exceptionally easy to use a centralized celery app).

I think that we should, however, continue to develop against the RESTful API - despite the fact that performance wise it makes more sense for the client to be using django models directly. The reason being it will force us to make the API robust and this will certainly be useful when extending the angular client.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ tags
 # Build byproducts
 docs/build
 /static
+ployst/settings/dev_al.py
--- a/docs/source/celery.rst
+++ b/docs/source/celery.rst
@@ -19,7 +19,7 @@ You should see log messages in the celery worker console.
 Adding celery tasks
 ===================
 
-Celery tasks are found if they are put inside a ``tasks.py`` modue
+Celery tasks are found if they are put inside a ``tasks.py`` module
 of an app that is registered in ``INSTALLED_APPS`` of settings.
 
 
--- a/docs/source/celery.rst
+++ b/docs/source/celery.rst
@@ -1,0 +1,25 @@
+Running celery worker
+=====================
+
+1. Make sure rabbitmq-server is running locally::
+
+    sudo rabbitmq-server
+
+2. Run the celery worker::
+
+    celery worker --app=ployst -l info
+
+3. To test it has worked, run the github_poke::
+
+    python manage.py github_poke http://github.com/pretenders/ployst develop
+
+You should see log messages in the celery worker console.
+
+
+Adding celery tasks
+===================
+
+Celery tasks are found if they are put inside a ``tasks.py`` modue
+of an app that is registered in ``INSTALLED_APPS`` of settings.
+
+
--- a/manage.py
+++ b/manage.py
@@ -3,7 +3,12 @@ import os
 import sys
 
 if __name__ == "__main__":
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "ployst.settings.dev")
+    if 'test' in sys.argv:
+        settings_path = 'ployst.settings.test'
+    else:
+        settings_path = 'ployst.settings.dev'
+
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", settings_path)
 
     from django.core.management import execute_from_command_line
 
--- a/ployst/__init__.py
+++ b/ployst/__init__.py
@@ -1,0 +1,3 @@
+from __future__ import absolute_import
+
+from .celery import app as celery_app  # NOQA
--- a/ployst/__init__.py
+++ b/ployst/__init__.py
@@ -1,3 +1,1 @@
-from __future__ import absolute_import
-
 from .celery import app as celery_app  # NOQA
--- a/ployst/apibase/management/commands/github_poke.py
+++ b/ployst/apibase/management/commands/github_poke.py
@@ -1,4 +1,5 @@
 import json
+from optparse import make_option
 
 import requests
 
@@ -7,6 +8,13 @@ from django.core.management.base import BaseCommand, CommandError
 
 class Command(BaseCommand):
     args = '<repo_url> <branch_name>'
+    option_list = BaseCommand.option_list + (
+        make_option('-p', '--port',
+            dest="port",
+            help="port to POST to (default: 8000)",
+            default='8000'
+            ),
+        )
     help = """Create a fake message from github to force an update.
     Requires a running instance"""
 
@@ -16,8 +24,9 @@ class Command(BaseCommand):
                 self.args, len(args)))
 
         repo_url, branch_name = args
-
-        url = 'http://localhost:8000/providers/github/receive-hook/tOkEn/'
+        port = options.get('port')
+        url = ('http://localhost:{port}/'
+               'providers/github/receive-hook/tOkEn/'.format(port=port))
         load = json.dumps({
             'repository': {
                 'url': repo_url
--- a/ployst/apibase/management/commands/github_poke.py
+++ b/ployst/apibase/management/commands/github_poke.py
@@ -1,0 +1,31 @@
+import json
+
+import requests
+
+from django.core.management.base import BaseCommand, CommandError
+
+
+class Command(BaseCommand):
+    args = '<repo_url> <branch_name>'
+    help = """Create a fake message from github to force an update.
+    Requires a running instance"""
+
+    def handle(self, *args, **options):
+        if len(args) != 2:
+            raise CommandError("Expecting 2 arguments: {}. Got {}".format(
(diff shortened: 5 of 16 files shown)


alexcouper:
On reflection, this isn't so controversial after all.

If we ever want to run the providers separately they will need to be run with a settings file that names them in installed apps, and have their own celery runner. But these are reasonable constraints to have.

alexcouper:
@txels when you get a moment ;)

txels:
Acknowledged. Will review shortly.
I don't think there is much controversy really, seems like a reasonable approach.

docs/source/celery.rst
```diff
@@ -0,0 +1,25 @@
...
+
+    python manage.py github_poke http://github.com/pretenders/ployst develop
+
+You should see log messages in the celery worker console.
+
+
+Adding celery tasks
+===================
+
+Celery tasks are found if they are put inside a ``tasks.py`` modue
```
txels: s/modue/module

manage.py
```diff
@@ -3,7 +3,12 @@
 import sys
 
 if __name__ == "__main__":
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "ployst.settings.dev")
+    if 'test' in sys.argv:
+        settings_path = 'ployst.settings.test'
+    else:
+        settings_path = 'ployst.settings.dev'
```
txels, line 9: sensible

ployst/__init__.py
```diff
@@ -0,0 +1,3 @@
+from __future__ import absolute_import
```
txels: do you need this here?

ployst/apibase/management/commands/github_poke.py
```diff
@@ -0,0 +1,31 @@
...
+    Requires a running instance"""
+
+    def handle(self, *args, **options):
+        if len(args) != 2:
+            raise CommandError("Expecting 2 arguments: {}. Got {}".format(
+                self.args, len(args)))
+
+        repo_url, branch_name = args
+
+        url = 'http://localhost:8000/providers/github/receive-hook/tOkEn/'
```
txels: port as an argument? I tend to use other ports for running

ployst/celery.py
```diff
@@ -0,0 +1,17 @@
...
+
+# set the default Django settings module for the 'celery' program.
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'ployst.settings.dev')
+
+app = Celery('ployst')
+
+# Using a string here means the worker will not have to
+# pickle the object when using Windows.
+app.config_from_object('django.conf:settings')
+app.autodiscover_tasks(lambda: settings.INSTALLED_APPS)
```
txels, line 17: interesting, I had never seen celery used without some `django-celery`.

docs/source/celery.rst
```diff
@@ -0,0 +1,25 @@
+Running celery worker
+=====================
+
+1. Make sure rabbitmq-server is running locally::
+
+    sudo rabbitmq-server
```
txels, line 6: would be nice to start adding system package dependencies somewhere in requirements

txels, line 6: BTW do you use vagrant?

docs/source/celery.rst
```diff
@@ -0,0 +1,25 @@
...
+
+2. Run the celery worker::
+
+    celery worker --app=ployst -l info
+
+3. To test it has worked, run the github_poke::
+
+    python manage.py github_poke http://github.com/pretenders/ployst develop
+
+You should see log messages in the celery worker console.
```
txels, line 16: nice useful doc.

ployst/github/tasks.py
```diff
@@ -0,0 +1,87 @@
...
+
+        controller = HierarchyHandler(path=repo['local_path'])
+
+        for feature in matched_features:
+            hierarchy = controller.get_branch_hierarchy(
+                feature['feature_id'],
+                regexes
+            )
+            statuses = controller.get_branch_merge_statuses(
+                hierarchy, branch_name
```
txels, line 65: this fails if I run the poke for `master`. It calls is_contained with a branch_name of `None`.

docs/source/celery.rst
txels, line 16: One thing I had to figure out by myself was what to add to provider settings. I added this:

```
{"branch_finders": ["^master$", "^develop$", ".*(?i){feature_id}.*"]}
```

I had a ployst repo and a feature 312 in my DB and tested with `python manage.py github_poke http://github.com/pretenders/ployst carles/312-api-security`

In order to debug with breakpoints I had to make `recalculate` sync by removing `.delay`. Is there some setting that can be used just for that (ignore celery and call the function synchronously)?

txels:
It all looks good to me. I added some comment/question, happy to merge on pingback.
BTW there is a GITHUB_HOOK_TOKEN defined on settings.dev - this is probably your local token? It is not the one I have now, so I forgot when that was added there. This should probably go in personal settings.

e.g. I have now ployst/settings/dev_cba.py:

``` python
from .dev import *  # noqa
GITHUB_HOOK_TOKEN = "27665cd2-c386-4fa6-a741-656fcf23ca65"
```

alexcouper:
Yeah - personal settings is where that should live. I must have committed by mistake.

docs/source/celery.rst
alexcouper, line 16: `CELERY_ALWAYS_EAGER = True` is what you're after.

It's what is in `settings/test.py` so we can write tests that run without the celery worker

docs/source/celery.rst
alexcouper, line 6: No actually. Since starting at my recent work place I've switched to developing on Mac only.

ployst/celery.py
alexcouper, line 17: In the latest celery it works out of the box. This was largely taken from their site.

ployst/__init__.py
alexcouper: I think not. Removed

ployst/apibase/management/commands/github_poke.py
alexcouper: Added optional arg.

docs/source/celery.rst
alexcouper: fixed.

alexcouper:
Remind me how you're getting your personal settings to be used?

Are you doing:
`DJANGO_SETTINGS_MODULE=ployst.settings.dev_cba python manage.py runserver`

docs/source/celery.rst
alexcouper, line 16: Turns out for debugging having `CELERY_EAGER_PROPAGATES_EXCEPTIONS = True` is also a good idea.

ployst/github/tasks.py
alexcouper, line 65: This has been fixed now.

txels:
Yes either that or `export DJANGO_SETTINGS_MODULE=ployst.settings.dev_cba` at the start (e.g. you can use a virtualenvwrapper post-activate hook for that (some colleagues do) I then to have a script `mysettings.sh` that I run `. ./mysettings.sh` when I need to.

ployst/github/views.py
```diff
@@ -108,6 +25,6 @@ def receive_hook(request, hook_token):
         LOGGER.error('Unexpected data structure: {0}'.format(request.POST))
         return HttpResponseBadRequest()
 
-    # TODO: Make this an async task.
-    recalculate(url, branch_name)
+    f = recalculate.delay(url, branch_name)
+    print f
```
txels, line 29: :)
Maybe we want to introduce `logging` now?

ployst/celery.py
txels, line 17: nice

txels:
Fine to merge. You can decide when to remove that `print` statement.

.gitignore
```diff
@@ -11,3 +11,4 @@ tags
 # Build byproducts
 docs/build
 /static
+ployst/settings/dev_al.py
```
txels, line 14: You can safely add those to version control. I'm adding mine.

PR #24, opened 2014-03-04
Additional user account views and styling

txels:
This is a follow-up to my former PR, includes a few more views and templates for user account management (slowly integrating `django-user-accounts`). I tried to use a more uniform DRY approach to styling.

Password reset doesn't work on accounts that have been created manually. I guess one has to go through the app's registration process.

Now that I've started using it I'm of two minds about keeping using this app or going for https://django-registration.readthedocs.org/en/latest/quickstart.html which I had used in the past. I can't see clear advantages to django-user-accounts and the fact that it apparently only fully works for accounts registered via itself is a bit annoying.


alexcouper:
I'm happy to merge this - or would you rather I didn't if you're going to move to django-registration?

txels:
Happy either way.
I'll rebase and re-push maybe with django-registration maybe not.

The main distinguishing point with django-user-accounts seems to be to have multiple email addresses for one user for login: http://django-user-accounts.readthedocs.org/en/latest/faq.html#why-can-email-addresses-get-out-of-sync - do we think we need/want that? I think we can do without. (We can always have multiple notification addresses à la github later on if we need that.)

Thoughts?

alexcouper:
I'm happy to merge in that case. Although you'll need to pull in from develop ;)

txels:
I've gone for django-registration in the end, although for now only `django.contrib.auth` views have been integrated (no registration workflow yet).

Please test that everything works for you (login, logout, password reset...) and I haven't forgotten to include any template in version control...

alexcouper:
It works fine - the password reset form seems to be in admin style rather than following the same UI as the rest. Is that expected for now?

txels:
No. I probably forgot to add some template.

txels:
OK should be fine now, I forgot to include a couple of templates.

ployst/urls.py
```diff
@@ -11,7 +13,11 @@
 
     # django and 3rd party apps
     url(r'^admin/', include(admin.site.urls)),
-    url(r"^account/", include("account.urls")),
+
+    #url(r'^account/login/', LoginView.as_view(), name='account_login'),
+    #url(r'^account/', include('account.urls')),
```
alexcouper: remove completely?

alexcouper:
When I click on the `Sign into your Ployst account` button i get an error in the backend:

```
WARNING:root:Could not resolve form field 'remember'.
  File "/Users/alex/.virtualenvs/ployst/lib/python2.7/site-packages/crispy_forms/utils.py", line 74, in render_field
    field_instance = form.fields[field]
KeyError: u'remember'
```

alexcouper:
Yep, the form looks fine now.

txels:
Oh that field was probably only available for the old accounts app. I'll check.

txels:
Should be solved now. I may later add some tests for regression.

alexcouper:
Hmm it would appear that this may have broken tests...

Not sure how yet. I'm getting:

`OperationalError: no such table: account_account`

when trying to run the tests...

alexcouper:
Turned out that I needed to uninstall django-user-accounts.

Plus remove some middleware (https://github.com/pretenders/ployst/pull/27)

PR #28, opened 2014-03-07
github hook token

alexcouper:
This includes:
- Rename the existing HOOK_TOKEN to be CORE_API_TOKEN since that is what we're actually using it for.
- Compare the hook_token received from github with a hash of the repo url and see that they match.

Once this has been merged (assuming #27 has been merged) all tests pass.


ployst/github/views.py
```diff
@@ -25,6 +26,19 @@ def receive_hook(request, hook_token):
...
+
+def create_token(repo_url):
+    """
+    Create a token for the given repo.
+
+    This token will be used in the url of requests from Github.
+    We will then compare it with the repo information contained in the body
+    of the request and if they match then it is deemed valid.
+    """
+    return hashlib.md5(repo_url + settings.GITHUB_HOOK_TOKEN_SALT).hexdigest()
```
txels, line 44: sha1 or sha256 are safer than md5.
Not that this is military grade ;)

ployst/settings/dev_al.py
```diff
@@ -0,0 +1,5 @@
+from .dev import *  # noqa
+
+GITHUB_CORE_API_TOKEN = "ba6d427e-a893-4014-9633-e38c55210b52"
+CELERY_ALWAYS_EAGER = True
+CELERY_EAGER_PROPAGATES_EXCEPTIONS = True
```
txels, line 5: should this one be in `dev.py`?

ployst/github/test/test_receive_hook.py
```diff
@@ -9,9 +11,21 @@
...
 
-    def post(self, data):
+    def _get_hook_token(self, data):
+        "Generate a token for the data provided"
+        try:
+            payload = json.loads(data['payload'])
+            url = payload['repository']['url']
+            return views.create_token(url)
+        except (KeyError, ValueError):
+            return ""
```
txels, line 21: not sure why you need try except here. We'll only call from appropriate test cases.

txels:
Yay green!

ployst/settings/dev_al.py
alexcouper, line 5: Probably worth while.

ployst/github/test/test_receive_hook.py
alexcouper, line 21: Yeah, the tests would need a slight refactor to explicitly generate a hook token or not. Probably worth it though.

PR #29, opened 2014-03-08
Front-end bits and pieces

txels:
As I was preparing to starting adding in Angular JS stuff:
- Implements an earlier proposal of yours to extend the `fab develop` task to install all dependencies
- Sets the right (latest stable) versions of angularjs
- Brings back the old `deploystream` "marketing" page. Not sure where this will end up living (this kind of stuff works better as pre-compiled static pages further down the line) but it will serve for a while. Choice of colors, graphics, background photos etc will come at some point.


fabfile.py
```diff
@@ -27,6 +29,21 @@ def develop():
     Update environment for local development
```
alexcouper, line 29: ooh, i should add the linking of post-commit to this task.

PR #30, opened 2014-03-09
Finally some angular.

txels:
This may complete task http://txels.tpondemand.com/entity/289 which is the last remaining task in the initial setup story, plus a little more.

Includes some basic angular scaffolding, and a couple of initial controllers:
- to show crude profile information of logged-in user.
- to show teams and projects for logged-in user

In order to accomplish the latter I've added ownership filters to existing viewsets that showed teams and projects. I reviewed failing tests to use a token in order to bypass ownership checks.

### DIFF
--- a/ployst/templates/base.html
+++ b/ployst/templates/base.html
@@ -19,19 +19,12 @@
         <link href="/static/lib/font-awesome/css/font-awesome.min.css" rel="stylesheet" media="screen">
         <link href="//netdna.bootstrapcdn.com/font-awesome/4.0.3/css/font-awesome.min.css" rel="stylesheet">
 
-        <!-- Bootstrap { -->
+        <!-- JS required by Bootstrap { -->
         <script src="http://code.jquery.com/jquery.min.js"></script>
         <script src="/static/lib/bootstrap/dist/js/bootstrap.min.js"></script>
         <!-- } -->
 
-        <style>
-            [ng\:cloak], [ng-cloak], [data-ng-cloak], [x-ng-cloak], .ng-cloak, .x-ng-cloak {
-                display: none;
-            }
-        </style>
-
-        {% block extrahead %}
-        {% endblock %}
+        {% block extrahead %}{% endblock %}
     {% endblock %}
     </head>
 
@@ -42,7 +35,7 @@
             </header>
 
             {# Main area where angular views will display #}
-            <section id="content" class="container-fluid" ng-view>
+            <section id="content" class="container-fluid"{% block ng_attrs %}{% endblock %}>
                 {% block content %}{% endblock %}
             </section>
 
@@ -54,14 +47,6 @@
             <p><small>&copy; 2013-2014 Alex Couper and Carles Barrob&eacute;s.</small></p>
             </div>
         </footer>
-
-        <script src="/static/lib/angular/angular.js"></script>
-        <script src="/static/js/app.js"></script>
-        <script src="/static/js/services.js"></script>
-        <script src="/static/js/controllers.js"></script>
-        <script src="/static/js/filters.js"></script>
-        <script src="/static/js/directives.js"></script>
-        <script src="/static/lib/angular/angular-resource.js"></script>
 
         {% block extrafoot %}{% endblock %}
     </body>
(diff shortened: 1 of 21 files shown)


alexcouper:
After I've logged in, the admin button no longer seems to work. I've tried removing the cache.
screen shot 2014-03-11 at 9 41 32 pm

ployst/templates/base.html
```diff
@@ -42,27 +36,17 @@
...
-        <footer id="footer">
+        <footer id="footer" class="container-fluid">
             <div class="container-fluid">
             <p>Produced by <a href="http://github.com/pretenders/">Pretenders</a>.</p>
             <p><small>&copy; 2013-2014 Alex Couper and Carles Barrob&eacute;s.</small></p>
             </div>
         </footer>
 
-        <script src="/static/lib/angular/angular.js"></script>
-        <script src="/static/js/app.js"></script>
```
alexcouper, line 59: How are all of these things loaded now?

ployst/ui/static/js/ployst.ui.admin.js
```diff
@@ -0,0 +1,82 @@
...
+
+    var ng = ployst.ng;
+
+    ng.config = ng.config || {};
+
+    ng.config.routing = function ($routeProvider, $locationProvider) {
+        $routeProvider
+            .when('/profile', {
+                controller: 'profile',
+                templateUrl: STATIC_URL + 'templates/profile.html',
```
alexcouper, line 13: I'm surprised that `STATIC_URL` is available in this context, i'd expect `window.STATIC_URL`.

ployst/ui/static/js/ployst.ui.admin.js
```diff
@@ -0,0 +1,82 @@
+(function () {
```
alexcouper, line 1: What's the thought behind the naming scheme of the JS modules for angular?

Eg, why `ployst.ui.admin.js` (for my benefit when moving forward)

ployst/templates/base.html
txels, line 59: The angular libs are in `base-angular.html`.
The app-specific scripts are loaded in `admin_base.html` which contains the "admin" app (this can do with a better name, but I am not sure about the UI we want to build "as a whole"). The structure has been simplified for now to just two files.

ployst/ui/static/js/ployst.ui.admin.js
txels, line 13: That is javascript in a browser for you. In fact, `window` is the global variable context, so `window.STATIC_URL` and `STATIC_URL` are the same. I got into the convention of my colleague front-enders to assign global variables in `window` so that it doesn't like you forgot to add the `var` declaration (that would make it a local variable).

I didn't add `window.` here because it makes the code look a bit cumbersome, but we could if you think it's clearer.

ployst/ui/static/js/ployst.ui.admin.js
txels, line 1: I'm still looking for proper names and breakdown. It will depend how we want to tie up the whole UI in terms of angular "applications". I'd like to have different apps for different parts of the UI. The `admin` app is what I've called this account/profile management side. I think the "storyboard" types of views will live in a different _ng-app_. But I'm not 100% sure TBH. If this app grows large it could be split into controllers, resources, etc.

The `ng` file contains all the "angular-y" declarations. This is something I am copying from the approach of my JS colleagues. I think it gives you a good view of the app skeleton in one single not too large file. It could be renamed `ployst.ui.admin-ng.js` or similar for clarity.

ployst/ui/static/templates/teams.html
```diff
@@ -0,0 +1,15 @@
+<h3><i class="fa fa-group"></i> Your Teams</h3>
+
+<ul>
+    <li ng-repeat="team in teams">
+        {{ team.name }} <kbd>{{ team.guid }}</kbd>
```
alexcouper, line 5: Was the `team.guid` meant to still be here?

txels, line 5: I added this to have something to show for teams while testing the UI. I was thinking of showing team members in here somehow.

This is all just a way to get the angular pieces working and show something but we need to discuss UI sometime soon.

ployst/ui/static/js/ployst.ui.admin.js
alexcouper, line 13: no - it's fine. It was a more a remark

PR #31, opened 2014-03-11
Initial heroku work.

alexcouper:
Seems to be working now.
@gissues:{"order":50,"status":"backlog"}


requirements.txt
```diff
@@ -0,0 +1,53 @@
+Django==1.6.1
```
alexcouper: Annoyingly, this didn't seem to work just having `-r requirements/heroku.txt`

ployst/wsgi.py
```diff
@@ -8,7 +8,15 @@
 """
 
 import os
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "ployst.settings")
+if 'ON_HEROKU' in os.environ:
```
alexcouper: Suggestions on how these can be shared between wsgi.py, celery.py and manage.py are welcome!

fabfile.py
```diff
@@ -47,3 +50,34 @@ def develop():
...
+@task
+def heroku_package_npm():
+    """
+    Create package.json for heroku
+    """
+    if not exists('package.json'):
+        local('npm init')
+    install_from_file('npm install', 'npm-modules', '@',
+                      command_opts='--save')
+
```
txels, line 64: pep8 will require an extra blank line

fabfile.py
```diff
@@ -47,3 +50,34 @@ def develop():
...
         local('bower install')
+
+
+@task
+def heroku_package_npm():
+    """
+    Create package.json for heroku
+    """
+    if not exists('package.json'):
+        local('npm init')
```
txels, line 61: maybe we want to use that same approach on our local set up?

package.json
```diff
@@ -0,0 +1,40 @@
...
+  "author": "",
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/pretenders/ployst/issues"
+  },
+  "homepage": "https://github.com/pretenders/ployst",
+  "dependencies": {
+    "bower": "^1.2.8",
+    "less": "^1.4.2",
+    "karma-script-launcher": "^0.1.0",
```
txels, line 22: this one and below should be `devDependencies` (not sure if it's written exactly like this)

ployst/celery.py
```diff
@@ -7,7 +7,12 @@
 from django.conf import settings
```
txels: this looks strange here with all the code below changing where settings are to be loaded from...

ployst/settings/heroku.py
```diff
@@ -0,0 +1,27 @@
...
+# Allow all host headers
+ALLOWED_HOSTS = ['*']
+
+# Static asset configuration
+import os
+STATIC_ROOT = 'staticfiles'
+STATIC_URL = '/static/'
+
+STATICFILES_DIRS = (
+    os.path.join(ROOT_DIR, 'assets'),
```
txels: isn't this in `base`? we should avoid duplicate settings to keep this minimal...

ployst/wsgi.py
txels: This looks extremely annoying.
Can we set the `DJANGO_SETTINGS_MODULE` environment variable in heroku? I understand you can set env vars in heroku arbitrarily? https://devcenter.heroku.com/articles/config-vars

requirements.txt
txels: we should then script the creation of this file, so that we can do it with a fab task as part of the heroku deploy and keep out of version control...

```
pip freeze > requirements.txt
cat requirements/heroku.txt >> requirements.txt  # and leave the -r bit out of heroku.txt.
```

...at the end of a successful build.

requirements/heroku.txt
```diff
@@ -0,0 +1,6 @@
+-r runtime.txt
```
txels: (this bit)

fabfile.py
```diff
@@ -47,3 +50,34 @@ def develop():
...
+
+
+@task
+def heroku_deploy():
+    """
+    Push this branch to heroku to deploy it.
+    """
+    local('heroku config:set ON_HEROKU=true')
+    this_branch = local("git rev-parse --abbrev-ref HEAD", capture=True)
+    local("git push heroku {0}:master".format(this_branch))
```
txels, line 85: would be nice to start making ourselves a collection of fab tasks à la `gdfabric`.

alexcouper, line 85: Yeah I think that is the way we want to go.

fabfile.py
alexcouper, line 64: sorted

package.json
alexcouper, line 22: Changed

ployst/celery.py
alexcouper: True. Moved down...

ployst/wsgi.py
alexcouper: Yeah good point - i've added that to the fab deploy script

requirements.txt
alexcouper: That would be ideal.

However, AFAIK the only way to deploy to heroku is by pushing a branch to it. Therefore the file has to be in version control.

alexcouper: I'm going to try a symlink...

alexcouper:
Should be more suitable now.

Once merged I'll redeploy ;)

requirements.txt
```diff
@@ -0,0 +1,8 @@
+#HEROKU REQUIREMENTS FILE
+
+-r requirements/runtime.txt
```
txels, line 3: oh so it did work?

PR #32, opened 2014-03-15
Admin UI - additional rework

txels:
Improved prototype more in the general direction of the layout I think we should go for.
Changes mainly in teams and projects page.


ployst/ui/templates/ui/admin_base.html
```diff
@@ -14,8 +14,11 @@
                 <a href="#/providers"><i class="fa fa-fw fa-gears"></i> Providers</a>
             </li -->
         </ul>
+        <footer id="mini-footer">
+            <p><small>&copy; 2013-2014 <a href="http://github.com/pretenders/">Pretenders</a></small></p>
+        </footer>
     </div>
-    <div class="col-sm-10" id="admin-main">
+    <div class="col-sm-offset-2 col-sm-8" id="admin-main">
```
alexcouper, line 21: Do we want to enforce this white space? On my browser I see a gap as a result of the `col-sm-offset-2` class.

screen shot 2014-03-15 at 10 15 08 pm

ployst/ui/templates/ui/admin_base.html
```diff
@@ -14,8 +14,11 @@
                 <a href="#/providers"><i class="fa fa-fw fa-gears"></i> Providers</a>
             </li -->
         </ul>
+        <footer id="mini-footer">
+            <p><small>&copy; 2013-2014 <a href="http://github.com/pretenders/">Pretenders</a></small></p>
```
alexcouper, line 18: On the home page we say `© 2013-2014 Alex Couper and Carles Barrobés.` which may be more legally correct.

We should probably be consistent one way or the other.

alexcouper:
This looks really good. I much prefer the page now that it doesn't have the bottom footer in it.

I think we may get a similar gain by making the top bar less "in your face".

ployst/ui/templates/ui/admin_base.html
txels, line 21: looks like you're not running with the latest CSS... can you try after collecstatic?

txels:
Yes I was planning to do the same with the top navbar as to the footer, once logged in.

ployst/ui/templates/ui/admin_base.html
txels, line 18: ha yes I did for it to take up less space but will use our names for now. I could get rid of it TBH.

ployst/ui/templates/ui/admin_base.html
alexcouper, line 21: ah yes, you suspected correctly.

PR #33, opened 2014-03-15
dj_static is not a requirement for local development

txels:
After the changes done for heroku, the local development environment was broken, i.e. even after running `fab develop`, `./manage.py runserver` failed. 

As `dj_static` is only a requirement for heroku, I've made sure `wsgi.py` is still usable outside heroku.


alexcouper:
Nice - this means you are deploying somewhere?

txels:
no, just trying to run locally.

PR #34, opened 2014-03-16
Use the correct JS file.

alexcouper:
This broke on heroku.


txels:
duh.

txels:
Looks like some sort of end to end testing (UI-level testing) is in order...

PR #35, opened 2014-03-17
Do not show top navigation in internal UI

txels:
No top nav once logged in.

I had to do some reshuffling of templates. I will probably do a little more clean up of template structure and naming sometime later.


assets/less/site.less
```diff
@@ -120,18 +128,43 @@ body {
...
     z-index: 110;
     width: 100%;
     background-color: @gray-dark;
+    padding: 6px;
+}
+
+#mini-header {
+    padding: 8px 16px 32px;
+    text-transform: lowercase;
+    #gradient.vertical(@gray-dark, @sidebar-background, 70%, 85%);
```
alexcouper, line 144: This looks like it's supposed to be commented out, but still renders the gradient in CSS.

I'm not a huge fan of the effect that it generates:
screen shot 2014-03-18 at 5 59 58 pm

But, not a blocker for the merge anyway...

txels, line 144: `#` is not a comment leader in CSS/LESS.
I can get rid of the gradient, I also don't particularly like the end result. I wanted some smooth visual transition from heading to menu.

PR #36, opened 2014-03-18
Set heroku config in one command.

alexcouper:
Soon I'll move these out into a fabric like package.

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -79,7 +79,6 @@ def heroku_deploy():
     """
     Push this branch to heroku to deploy it.
     """
-    local('heroku config:set ON_HEROKU=true')
-    local('heroku config:set DJANGO_SETTINGS_MODULE=ployst.settings.heroku')
+    local('heroku config:set ON_HEROKU=true DJANGO_SETTINGS_MODULE=ployst.settings.heroku')
     this_branch = local("git rev-parse --abbrev-ref HEAD", capture=True)
     local("git push heroku {0}:master".format(this_branch))
--- a/fabfile.py
+++ b/fabfile.py
@@ -72,13 +72,14 @@ def heroku_configure():
         'heroku config:add '
         'BUILDPACK_URL=https://github.com/ddollar/heroku-buildpack-multi.git'
     )
-
+    local('heroku config:set '
+          'ON_HEROKU=true '
+          'DJANGO_SETTINGS_MODULE=ployst.settings.heroku')
 
 @task
 def heroku_deploy():
     """
     Push this branch to heroku to deploy it.
     """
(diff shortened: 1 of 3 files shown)


txels:
BTW in terms of credentials, what do I need in order to run these?

Also I realised that `heroku` is nowhere in requirements. We should have a `requirements/deploy.txt` or `deploy-heroku.txt`?

alexcouper:
It's not something that I installed using pip. I got it here:

https://toolbelt.heroku.com/

I should probably add that to the docs...unless you know of a pipable alternative

txels:
No I don't. Yes I'd add to docs.

txels:
Cool. I can now push to heroku (I pushed develop). But with git push, the fabric task does not work:

```
$ fab heroku_deploy
[localhost] local: heroku config:set ON_HEROKU=true DJANGO_SETTINGS_MODULE=ployst.settings.heroku
 !    No app specified.
 !    Run this command from an app folder or specify which app to use with --app APP.
```

txels:
It works if I shift things around like this:

``` python
@task
def heroku_configure():
    """
    Configure for a heroku deploy
    """
    local(
        'heroku config:add '
        'BUILDPACK_URL=https://github.com/ddollar/heroku-buildpack-multi.git'
    )
    local('heroku config:set ON_HEROKU=true DJANGO_SETTINGS_MODULE=ployst.settings.heroku')


@task
def heroku_deploy():
    """
    Push this branch to heroku to deploy it.
    """
    this_branch = local("git rev-parse --abbrev-ref HEAD", capture=True)
    local("git push heroku {0}:master".format(this_branch))
```

I.e. keep config separate from deploy. I cannot run config though for the same reason above. I'm guessing you have some local heroku config somewhere...

alexcouper:
Strange, I wonder why.
Does the configure command work for you after the reshuffle?

There may be some heroku settings we are missing...

txels:
No, heroku config commands don't work for me. It can't figure out the app.

fabfile.py
```diff
@@ -79,7 +79,6 @@ def heroku_deploy():
     """
     Push this branch to heroku to deploy it.
     """
-    local('heroku config:set ON_HEROKU=true')
-    local('heroku config:set DJANGO_SETTINGS_MODULE=ployst.settings.heroku')
+    local('heroku config:set ON_HEROKU=true DJANGO_SETTINGS_MODULE=ployst.settings.heroku')
```
txels: By looking at the log on heroku I assume we don't need to do this every single time we deploy. It counts this as a config change and as such an extra application version.

alexcouper: ok, i'll move that out...

alexcouper:
I think this is ready for a another glance at now...

PR #37, opened 2014-03-18
Carles: start story 324 manage teams

txels:
Story: http://txels.tpondemand.com/entity/324

This is a bit of a mix but as I was doing some work on http://txels.tpondemand.com/entity/324 I included your suggestion to get rid of the gradient as well.

This is a first not very styled version of being able to delete projects and teams in the main UI. Next steps for this story are going to be create projects and teams, and invite users to a team or remove them from one.

### DIFF
--- a/ployst/settings/base.py
+++ b/ployst/settings/base.py
@@ -155,10 +155,6 @@ COMPRESS_PRECOMPILERS = (
     ('text/less', LESS_COMMAND),
 )
 
-REST_FRAMEWORK = {
-    'DEFAULT_FILTER_BACKENDS': ('rest_framework.filters.DjangoFilterBackend',)
-}
-
 LOGGING = {
     'version': 1,
     'disable_existing_loggers': False,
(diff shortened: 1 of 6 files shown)


alexcouper:
Looks very elegant.

ployst/settings/base.py
```diff
@@ -197,9 +193,15 @@
     },
 }
 
+# REST API ----------------------------------------------------------------
+REST_FRAMEWORK = {
+    'DEFAULT_FILTER_BACKENDS': ('rest_framework.filters.DjangoFilterBackend',)
+}
+APPEND_SLASH = False  # play nicely with angularJS no end slashes
```
alexcouper, line 200: Why does angularJS not like end slashes?

txels, line 200: Angular JS for some reason removes end slashes to normalise URLs (the opposite to default Django). There are lots of blog posts mostly from django people complaining about this mismatch. 

This was the cleanest solution I found on one of those.

txels, line 200: BTW I had to clear the cache of my browser at some point as Django does the / appending as 301 (permanent redirects).

PR #38, opened 2014-03-22
Carles/324 manage teams

txels:
This is a more fully working prototype for http://txels.tpondemand.com/entity/324

It's prototypal as I'm still figuring out how to structure things in Angular, markup, LESS and DRF (lots of open fronts) so there's loads of untested code. But the basic UI works, have a play with it and tell me what you think.

All the forms that are visible are meant to be hidden and open up in pop-overs when clicking on some sort of (+) or (Add) button.


ployst/ui/static/templates/teams.html
```diff
@@ -1,31 +1,66 @@
...
+            <li ng-repeat="user in team.users">
+                <i ng-if="isManager(team, user)"
+                    class="fa fa-fw fa-star" title="manager"></i>
+                <i ng-if="!isManager(team, user)"
+                    class="fa fa-fw"></i>
+                <span>{{ user.username }}</span>
+            </li>
+            <li ng-if="isManager(team, user)">
+                <h5>Invite a user to join {{ team.name }}:</h5>
+                <input ng-model="newUser.email" name="email" placeholder="email..."></input>
```
alexcouper, line 36: For now, maybe replace email with "ployst username" as no invite is sent at the moment?

alexcouper:
Works well!

alexcouper:
I found the position of  "add new team"  to be confusing, in the sense that I wondered whether I was adding a team within a team due to the layout of the page. Perhaps it should be along the top with the teams?

screen shot 2014-03-23 at 10 42 14 am

alexcouper:
Or, alternatively, it might be a case of just visually adjusting both the delete and add buttons.

alexcouper:
Happy to merge. Let me know if you want it in develop as is...

ployst/core/accounts/views.py
```diff
@@ -30,11 +34,72 @@ class ProjectViewSet(PermissionsViewSetMixin, ModelViewSet):
...
+
+        """
+        super(TeamViewSet, self).create(request, *args, **kwargs)
+        TeamUser.objects.create(
+            team=self.object, user=request.user, manager=True
+        )
+        serializer = self.get_serializer(self.object)
+        return Response(serializer.data)
+
+    @action()
```
alexcouper, line 68: This is pretty cool

ployst/core/accounts/views.py
```diff
@@ -30,11 +34,72 @@ class ProjectViewSet(PermissionsViewSetMixin, ModelViewSet):
...
+    def invite_user(self, request, pk=None):
```
alexcouper, line 69: As you say, this could do with tests...

ployst/ui/static/templates/teams.html
txels, line 36: I thought about that, and I still think people are more likely to know friends/colleagues by email than ployst username. Although we could easily accept either.

ployst/core/accounts/views.py
txels, line 68: yes, DRF has quite a few nifty features.

txels:
I've moved team creation "form" to the top, outside individual team section, as per your suggestion.

Re this PR I'd like your general opinion on a few things:
- How the code is structured (`ui` app vs alternatively moving this UI to the accounts app, as everything so far is about managing account data).
- This code will not work if an email address is used by more than one account. Do we want to enforce that uniqueness, or should we support multiple accounts with the same email?
-  Do you think the _invitation_ approach hinted here is the way to go (even if it's not implemented until we include user signup)?

Re: merging, unless you want to use this to build something on top, I'd rather at least add some tests.

ployst/core/accounts/views.py
txels, line 69: I've now added some backend API tests.

txels:
One can now use `tox` to run a full build. It generates coverage data.
`flake8` fails but you fixed that on your GH oauth branch.

txels:
Now I merged your branch and fixed a test, and we have a green `tox` build :)

alexcouper:
I think we're missing tox as a dependency somewhere...

alexcouper:
Re your questions:
- How the code is structured (ui app vs alternatively moving this UI to the accounts app, as everything so far is about managing account data). 

I'm not sure I follow. It all looks like it is in the accounts app...
- This code will not work if an email address is used by more than one account. Do we want to enforce that uniqueness, or should we support multiple accounts with the same email?

I don't think we should support using the same email address on multiple accounts.
- Do you think the invitation approach hinted here is the way to go (even if it's not implemented until we include user signup)?

Yep - I think this makes sense. On the down side, I can forsee a g+ style annoyance when you don't know what email address a colleague signed up with and so you invite them again under a new email address and they end up with multiple accounts...

txels:
Re: g+ style annoyance, it's a good point, we should make sure our invitation process covers that case and users can accept an invitation to join a team from an existing account even if the email address doesn't match. Maybe, if we want to make it secure, with a counter-invitation or moderated acceptance?

PR #42, opened 2014-03-30
Alex/326 github oauth

alexcouper:
Here are the changes for the PR I had earlier.

There should be 1 failing test after this is merged: due to `NotImplementedError`


ployst/github/test/test_oauth.py
```diff
@@ -43,7 +43,7 @@ def test_handle_callback_from_github(self, oauth_exchange):
             }
         )
 
-        self.assertEquals(response.status_code, 200)
+        self.assertEquals(response.status_code, 302)
```
txels, line 48: :) I had fixed this already in my other PR: https://github.com/pretenders/ployst/pull/38/files#diff-c9d6fb42301f34c5fd520bbe40d6b352R47

ployst/github/__init__.py
```diff
@@ -2,4 +2,5 @@
 
 from ployst.core.client import Client
 
-client = Client('http://localhost:8000/', settings.GITHUB_CORE_API_TOKEN)
+client = Client(settings.GITHUB_CORE_API_ADDRESS,
```
txels: this setting is not github-specific?

txels:
Looks good to me.

ployst/github/views/oauth.py
```diff
@@ -20,7 +20,8 @@ def start(request):
     """
     return HttpResponseRedirect(
         'https://github.com/login/oauth/authorize?'
-        'client_id={0}&scope=repo,write:repo_hook&state={1}'.format(
+        'client_id={0}&scope=repo,write:repo_hook,write:public_key'
```
txels: Might be a good idea to keep a list of required github permissions in a central place? (one place could be in github/conf.py, although it's not really an overrideable setting, so it could as well be in `github/__init__` or `lib`).

``` python
GITHUB_OAUTH_SCOPE = ["repo", "write:repo_hook", "write:public_key"]
```

...and use `",".join(...)` here.

Would make future maintenance simpler.

ployst/github/__init__.py
alexcouper: Changed

alexcouper:
This now also includes checking flake8 before accepting commits.

PR #43, opened 2014-04-01
Alex/326 github oauth

alexcouper:
Add a `UserOAuthToken` and use it to store github tokens.


ployst/core/accounts/models.py
```diff
@@ -113,3 +113,13 @@ class ProjectProviderSettings(models.Model):
 
     class Meta:
         unique_together = ('project', 'provider')
+        verbose_name_plural = "ProjectProviderSettings"
```
txels, line 116: :) yess

alexcouper, line 116: Haha

ployst/core/client.py
```diff
@@ -101,5 +101,12 @@ def create_or_update_branch_information(self, branch_info):
         else:
             self.post('repos/branch/', branch_info)
 
-    def set_access_token(self, oauth_provider, access_token):
-        pass
+    def set_access_token(self, user_id, oauth_provider, access_token):
+        self.post(
+            'accounts/token',
```
txels, line 106: wondering whether we should use reverse here.

txels:
Looks good to me, let me give it a spin before merging.

ployst/core/client.py
```diff
@@ -101,5 +101,12 @@ def create_or_update_branch_information(self, branch_info):
...
+            data={
+                'user': user_id,
+                'identifier': oauth_provider,
+                'token': access_token,
+            },
+        )
```
txels, line 112: BTW this creates a new duplicate token every time I open `/providers/github/oauth-start`.
You probably need to override `create` on UserTokenViewSet to do get_or_create-ish

txels, line 112: This seems to work nicely:

``` python
class UserTokenViewSet(PermissionsViewSetMixin, ModelViewSet):
    model = UserOAuthToken

    def pre_save(self, instance):
        UserOAuthToken.objects.filter(
            user=instance.user, identifier=instance.identifier
        ).delete()
```

...deleting previous tokens from the same provider prior to saving.

https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/mixins.py#L49

txels:
Good stuff. Are you planning in using `github3` as github client as we did in deploystream? At this point I could try to use that to retrieve a list of repos and organisations to present the user with?

alexcouper:
Yeah `github3` was where I was thinking of going.

ployst/core/client.py
alexcouper, line 112: Feel free to push the change :)

ployst/core/client.py
alexcouper, line 106: I've stayed away from using reverse on the basis that this client will eventually live outside the django realm and won't have reverse capabilities there.

ployst/core/client.py
txels, line 112: Done, plus a test. I've also rebased.

txels:
Happy to merge if you are.

ployst/core/client.py
txels, line 106: Yes, I am thinking it would be good to have a way to generate some reversed URLs from core ployst that could be used from here and JS. Maybe an offline task that dumps some JSON?

txels:
I'm taking the liberty to merge this. Feel free to add comments post-merge to my additions.

PR #44, opened 2014-04-02
Carles/273 available providers

txels:
https://github.com/pretenders/ployst/pull/41 to develop


alexcouper:
Observations:
- I found I wanted to show the full first menu again after clicking on an item so I could be reminded what the headings were. The only way I could do this was to click on the profile button.
- Creation of a team felt a little out of place at the bottom of the sub menu - especially with a large gap between the previous entry and it.
- In normal operation these menu items would probably make sense to be within some kind of "configuration/settings" page. Are we thinking of having eg. "Providers" as a globally available shortcut? I would think you would set these up in the beginning and then rarely revisit. 

Merging as is

txels:
Cheers.

Re: your points:
1. True, there is no way to open it back. I was thinking of showing the icon titles on hover, or adding a back button in the submenu that showed back the main one. Ideas?
2. Yes I was thinking of showing just a plus button in the heading teams and the form open on popover (as a general approach to "add" forms elsewhere). I haven't gotten around to doing it yet. Thoughts?
3. In truth I was thinking of this we have now as the "settings" section. Normal operation for "story board" etc will look different I imagine (like e.g. github's account settings vs the rest).

txels:
BTW by focusing on the UI look we forgot to mention that this PR finally gets us to the point of addressing provider setup. Would you rather me start with github or TP?

alexcouper:
I'd be happy with either!

PR #49, opened 2014-04-11
Unit tests for teams controller

txels:
...more angular unit tests.

I've also added JS test coverage. Once tests are run (or they can be run continuously with `karma start`) an HTML coverage report is generated in `coverage/`.

### DIFF
--- a/aliases
+++ b/aliases
@@ -1,0 +1,3 @@
+alias bower=./node_modules/bower/bin/bower
+alias jshint=./node_modules/jshint/bin/jshint
+alias karma=./node_modules/karma-cli/bin/karma
--- a/requirements/npm-modules.txt
+++ b/requirements/npm-modules.txt
@@ -1,7 +1,0 @@
-bower==1.2.8           # front-end package manager
-less==1.4.2            # LESS/CSS compiler
-karma==0.12.6          # Javascript test runner
-karma-cli              # CLI for karma
-karma-coverage==0.2.1  # coverage reporter for karma
-phantomjs==1.9.1-0     # headless browser
-jshint==2.3.0          # JS code style check and lint
--- a/requirements/npm-modules.txt
+++ b/requirements/npm-modules.txt
@@ -1,5 +1,7 @@
 bower==1.2.8           # front-end package manager
 less==1.4.2            # LESS/CSS compiler
-karma==0.10.9          # Javascript test runner
+karma==0.12.6          # Javascript test runner
+karma-cli              # CLI for karma
+karma-coverage==0.2.1  # coverage reporter for karma
 phantomjs==1.9.1-0     # headless browser
 jshint==2.3.0          # JS code style check and lint
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 tags
 .ropeproject/
 /assets/lib
+/node_modules/
 
 # This may be a sublime thing
 .project
@@ -15,6 +16,7 @@ tags
 .coverage
 .tox
 /cover
+/coverage
 /docs/build
 /ployst.egg-info
 /static
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: python
 python: 2.7
 install:
   - pip install tox==1.7.1
-  - npm install -d
-  - npm install -g bower
-  - bower install
+  - npm install
 script:
   - tox
--- a/fabfile.py
+++ b/fabfile.py
@@ -1,29 +1,9 @@
 from os import environ
-from os.path import dirname, exists, join
+from os.path import dirname, exists
 from fabric.api import abort, local, task
-from fabric.contrib.console import confirm
 
 
 THIS_DIR = dirname(__file__)
-
-
-def install_from_file(command, filename, version_separator,
-                      command_opts=''):
-    """
-    Install requirement from a file with the given command.
-
-    The files use the standard version separator '==' which will get
-    replaced with whatever you pass in here in `version_separator`.
-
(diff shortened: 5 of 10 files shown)


alexcouper:
I think I'm being slow...but how do you get the coverage to work?

I ran `fab develop` and then `tox` and got coverage missing warnings.

Here's the output from running `karam start --single-run`

```
WARN [reporter]: Can not load "coverage", it is not registered!
  Perhaps you are missing some plugin?
INFO [karma]: Karma v0.10.10 server started at http://localhost:9876/
INFO [launcher]: Starting browser PhantomJS
WARN [preprocess]: Can not load "coverage", it is not registered!
  Perhaps you are missing some plugin?
INFO [PhantomJS 1.9.7 (Mac OS X)]: Connected on socket KnlcNA67urpx9Ixl42fB
PhantomJS 1.9.7 (Mac OS X): Executed 6 of 6 SUCCESS (0.121 secs / 0.037 secs)
```

txels:
For the time being `tox` does not install npm requirements within its virtual env and relies on global installs. Whereas `fab develop` installs these in your _local npm_ in the virtualenv. I have yet to update package.json for the travis build to work.

Check whether you are actually running the `karma` binary in your virtualenv (apparently not, as you are running an older version than the one installed by `fab develop`):

``` bash
$ workon ployst   # that's my virtualenv's name
$ which karma
/home/carles/.virtualenvs/ployst/bin/karma
$ karma start --single-run
WARN [karma]: Port 9876 in use
INFO [karma]: Karma v0.12.6 server started at http://localhost:9877/
```

requirements/npm-modules.txt
```diff
@@ -1,5 +1,7 @@
 bower==1.2.8           # front-end package manager
 less==1.4.2            # LESS/CSS compiler
-karma==0.10.9          # Javascript test runner
+karma==0.12.6          # Javascript test runner
+karma-cli              # CLI for karma
```
txels: newer versions of `karma` do not install a binary `karma` in the path, you need `karma-cli` for that.

alexcouper:
Looks like the `-g` in npm install screws with the upgrade. And unsurprisingly slow - it means install it globally - rather than in the VM!

Is it OK if we remove that `-g` option? 

```
(ployst)➜  ployst git:(carles/angular-tests) ✗ which karma
/Users/alex/.virtualenvs/ployst/bin/karma
(ployst)➜  ployst git:(carles/angular-tests) ✗ /Users/alex/.virtualenvs/ployst/bin/karma --version
Karma version: 0.10.10
(ployst)➜  ployst git:(carles/angular-tests) ✗ npm install -g karma@0.12.6
npm http GET https://registry.npmjs.org/karma/0.12.6
...
<Removed all the subsequent 304/GET requests>
...
npm http 304 https://registry.npmjs.org/options
(ployst)➜  ployst git:(carles/angular-tests) ✗ /Users/alex/.virtualenvs/ployst/bin/karma --version
Karma version: 0.10.10
```

Then, without the -g:

```
(ployst)➜  ployst git:(carles/angular-tests) ✗ /Users/alex/.virtualenvs/ployst/bin/karma --version
Karma version: 0.10.10
(ployst)➜  ployst git:(carles/angular-tests) ✗ npm install karma@0.12.6
npm http GET https://registry.npmjs.org/karma/0.12.6
...
<Removed all the subsequent 304/GET requests>
...
(ployst)➜  ployst git:(carles/angular-tests) ✗ /Users/alex/.virtualenvs/ployst/bin/karma --version
Karma version: 0.12.6
```

txels:
if you have run `nodeenv -p`, `which npm` should point to an npm version within your virtualenv. In that case, `npm install -g` installs packages within your virtualenv, and ensures binaries are installed in your virtualenv's `bin`.

`npm install` without -g, on the contrary, will just download in your working directory's `node_modules/` and won't care about ensuring scripts are in the path (bower, less, jshint, karma-cli... we rely on those commands being installed in the path, and some even complain if they are not `npm WARN prefer global jshint@2.3.0 should be installed with -g`). So they end up in places like `./node_modules/karma-cli/bin/karma`. We could just not care and always do a local install and always use that relative path. I will try that. It goes together with https://github.com/pretenders/ployst/pull/51/files but includes a little more work.

alexcouper:
```
(ployst)➜  ployst git:(alex/pre-commit) ✗ which npm
/Users/alex/.virtualenvs/ployst/bin/npm
```

I'm going to try and build the venv again from scratch.

txels:
I've simplified the whole npm dependency management process. Everything is now _npm-native_ (dependencies only in `package.json`) and relies on just `npm install`. You don't even need `nodeenv` now, just a decently up to date version of `node` (0.10) and `npm` (1.4) in your system. `tox` and `travis` are happy with the new setup. Builds are fast. `fab develop` also is.

On the flip side, all scripts are also installed only locally (in `node_modules` but not in the path - `jshint`, `bower`, `less`...) so invoking them is more cumbersome (e.g. `./node_modules/karma-cli/bin/karma`).

alexcouper:
Should we create some aliases?

alexcouper:
Oh you beat me to it!

aliases
```diff
@@ -0,0 +1,3 @@
+alias bower=./node_modules/bower/bin/bower
+alias jshint=./node_modules/jshint/bin/jshint
+alias karma=./node_modules/karma-cli/bin/karma
```
txels, line 3: these are just for my convenience.

alexcouper:
Ideally we'd ensure that we have node and npm available on your system and install if not available.

PR #62, opened 2014-07-19
Alex/github cloner

alexcouper:
- Generate a new key per repo.
- store that key next to the repo.


ployst/github/tasks/clone_repos.py
```diff
@@ -125,21 +105,29 @@ def ensure_clones_for_project(project_id):
     configured_repos = prov_settings['repositories']
```
txels: We should get the repos via the api from the project given.

alexcouper:
This is ready for another review.

ployst/core/urls.py
```diff
@@ -12,7 +12,7 @@
         include('ployst.core.builds.urls', namespace='builds')),
     url(r'^features/',
         include('ployst.core.features.urls', namespace='features')),
-    url(r'^providers',
+    url(r'^providers/',
```
txels: This breaks the angular API call to `"/core/providers"` (and runs into the issue that angular strips off trailing `/`).
This should have made a test fail, but it doesn't because we reverse URLs for testing.

ployst/github/tasks/clone_repos.py
```diff
@@ -122,24 +102,33 @@ def ensure_clones_for_project(project_id):
...
+    oauth_token = client.get_access_token(oauth_user_id, 'github')
 
     gh = github3.login(token=oauth_token)
-    for repo_path in configured_repos:
 
+    for repo in configured_repos:
+        repo_path = repo['path']
         destination = get_destination(repo_path)
-        if not os.path.exists(destination):
+        clone_location = join(destination, 'clone')
```
txels, line 93: Apparently we don't want to use `local_path` in the Repository model any longer, could delete it.

ployst/github/tasks/clone_repos.py
```diff
@@ -122,24 +102,33 @@ def ensure_clones_for_project(project_id):
...
+            clone_repo(repo, private, clone_location)
         # TODO: We need to now ensure that a repo.repository model instance
         # exists for the project in question. (cloned_already could be True
         # but from a different project.)
+
+
+if __name__ == '__main__':
+    # TODO
+    # Need to come up with a way to integration test this, rather than running
+    # it.
```
txels: Hm yes not sure how you test this, I run into various errors.

ployst/core/repos/models.py
```diff
@@ -43,6 +43,10 @@ class Repository(TeamObject):
     class Meta:
         verbose_name_plural = 'repositories'
 
+    @property
+    def path(self):
+        return self.url.split('http://github.com/')[-1]
```
txels: This isn't working for me, my repos have "https://github.com"... also trailing slash which didn't work.

Should we just stop storing the URL and store the path (or GH owner, name -which we already have- as two fields) and construct the URL when needed, with a property?

ployst/github/tasks/clone_repos.py
txels: I managed to run it from the shell, after manually entering some provider settings, copied from your test:

```
{
 "branch_finders": ["^master$", ".*(?i){feature_id}.*"],
 "repositories": ["pretenders/ployst"],
 "oauth_user": ["1"]
}
```

From what I've seen we no longer need repositories in here, they are entries in model `Repository` for the project.

In any case, no need to worry now, these settings will be migrated to be provider data at the project level.

ployst/github/test/__init__.py
```diff
@@ -1,12 +1,65 @@
...
+                "project": 1,
+                "team": "10123",
+                "local_path": DUMMY_REPO
+            },
+        ])
+
+    def get_provider_settings(self, project, provider_name):
+        return {
+            "branch_finders": ["^master$", ".*(?i){feature_id}.*"],
+            "repositories": ['pretenders/ployst'],
```
txels: apparently no longer needed

txels:
Gone through all, seems fine. I left some notes, not blockers.

ployst/github/tasks/clone_repos.py
alexcouper, line 93: Removed

ployst/github/test/__init__.py
alexcouper: Removed.

ployst/core/urls.py
alexcouper: I'm not sure if it has been released yet, but we can get round the angular stripping behaviour using:

https://github.com/angular/angular.js/pull/7064/files#diff-55d7a03dafb915a3531d8b59f34e3416R298

alexcouper: I've made the trailing slash optional for now

ployst/core/repos/models.py
alexcouper: I fixed the path property for now. We could change to constructing url from other fields.

alexcouper: Actually, thinking of going for your approach...

ployst/github/tasks/clone_repos.py
alexcouper: I'm going to add an "integration" folder and create a test in there that will require a backend running. I don't want it to be run from travis though, so i'll have to see what to do about that.

alexcouper:
This is done from my pov

ployst/core/repos/migrations/0004_auto__del_field_repository_url__add_field_repository_owner.py
```diff
@@ -0,0 +1,122 @@
+# -*- coding: utf-8 -*-
```
txels, line 1: Not long ago I learnt about `schemamigration --update`, that updates last migration instead of creating a new one.
How are we going to deal with migrations? Squash them often until we go live-ish?

ployst/github/management/commands/clone_repos.py
```diff
@@ -0,0 +1,88 @@
...
+from ployst.core.repos.models import Repository
+from ployst.github.views import oauth
+from ployst.github.tasks.clone_repos import ensure_clones_for_project
+
+from ployst.github import client
+
+
+class Command(BaseCommand):
+
+    help = """Check that end-to-end we can clone repos for a project
```
txels, line 24: Good solution for now

txels:
All good work. I'll run once locally and merge.

txels:
Kosher

PR #66, opened 2014-11-09
Getting rid of teams, code changes

txels:
I've reorganised some stories in TP according to our discussion of simplifying ployst and removing teams for now. This includes some code changes in JS and backend.

TODO:
- [x] redo the ownership checks/mixins to work with the new structure.
- [x] fix failing python tests
- [x] fix failing JS tests
- [x] write tests for the new projects code (JS in particular) which will in part be a port of the former team tests
- [x] leave the project detail page in some "finished" state (what is there, works - what doesn't work [yet] is not shown) - for now this means user management is very limited in functionality, but it works

Follow-up stories/work once this PR is done will include:
- Restructure the angular code. Break it down differently and into more files.
- Move the github-based repo selection (now in providers) to within projects, as discussed
- Reconsider the basic UI structure. I liked the approach of making good use of horizontal space by putting menus to the sides, but it calls for too much coding effort for custom UI components, so I'll bite the bullet and go for a canned framework. I want to try this: https://material.angularjs.org rather than just default to bootstrap3.


ployst/core/accounts/views.py
```diff
@@ -36,43 +36,22 @@ def retrieve(self, request, *args, **kwargs):
...
-            team=self.object, user=request.user, manager=True
-        )
-        serializer = self.get_serializer(self.object)
-        return Response(serializer.data)
-
     @action()
     def invite_user(self, request, pk=None):
         """
-        Invite a user to your team, from email address.
+        Invite a user to your project, from email address.
```
txels, line 55: Oh and we should also reconsider this, by letting you use GH handle or select users from your organisations. But that is the job for another PR.

txels:
@alexcouper this is now ready for a review and possible merge.
Test coverage may be sub-optimal, let me know if you identify some gap.

If you are happy to merge, before doing that you may consider looking at #68 and chain-merging

PR #67, opened 2014-11-14
Get rid of teams + ng restructure (combi PR)

txels:
This is a superset of https://github.com/pretenders/ployst/pull/66

It includes a major overhaul of the angular project structure, more in line with some best practices (e.g. [Google's](https://docs.google.com/a/gagnavarslan.is/document/d/1XXMvReO8-Awi1EZXAXS4PzDzdNvV6pGcuaF4Q9821Es/edit)). This will make the project more maintainable as it grows. It is only a first step, I haven't split things out like controllers and directives etc.

I've made it a separate PR because this is more unreadable than the other one. Feel free to close #66 if you prefer to merge this one directly.


ployst/ui/static/providers/ployst.providers.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * Ployst data providers management
+ *
+ * Enable and disable providers in ployst, and manage provider-specific
+ * settings.
+ */
+angular.module('ployst.providers', [
```
txels, line 7: We are not using this ATM, but it will become useful when we include things like global provider configuration. We might as well get rid of it for now though.

PR #71, opened 2014-11-20
Redirect to the correct location after oauth.

alexcouper:
This is a quick fix for an issue I saw that was after you oauth you were redirected to a different page than where the button was pressed.

This is a quick fix in the sense that it hard codes where you end up - if we want more places to enable oauth we will need to pass additional params to github that we can use when it comes back.


txels:
Ah I think we should in the longer term handle this via some special post-auth angular route that then redirects you to the right place (within the angular world).

PR #76, opened 2014-11-28
Create web hook from admin.

alexcouper:
This contains ability to create a web hook from the admin page.


alexcouper:
TODO
- [x] Add documentation about testing with pagekite
- [x] Add the secret when creating a hook and use it to verify authenticity of requests.
- [x] Check if a hook exists for this repo before adding a new one.

alexcouper:
@txels This is ready for review.

alexcouper:
Tested locally using pagekite

ployst/core/repos/admin.py
```diff
@@ -9,7 +9,21 @@ class BranchAdmin(admin.ModelAdmin):
 admin.site.register(Branch, BranchAdmin)
 
 
+def create_hooks(modeladmin, request, queryset):
+    """
+    Create hooks for the repositories.
+
+    TODO: For testing purposes only - this should be removed when we have this
+    hooked up to the front end, as we are importing github stuff into core.
```
txels, line 17: :) spooky
Could we do this in the github app `admin` somehow?

txels:
While testing the code in this PR (which works nicely) I was debugging the hook and realised that it's not really working:
- Our way of fetching repos by URL no longer works. I have replaced it locally with fetching by (owner, name), which does work.
- The `recalculate` algorithm is based on the existance of features, but doesn't update any info on local branches if features are missing - I suggest we implemented that to _close one loop_: update local branch info with updated branch data coming from the hook

txels:
I will push my fix to the first issue (fetch repo by URL).
Maybe you can work on the second one @alexcouper ?

alexcouper:
Re: second issue - just to be clear: we just want all branches to be stored in the DB for now, regardless of name etc. Is that right?

txels:
I think so. So that the current UI is self-contained. The project provider settings (branch patterns) cannot be set via the UI, and features do not exist yet. So the only MVP is keep branch info up to date with what we get from the hook.

alexcouper:
Second issue raised in #83

PR #79, opened 2014-11-29
Repo selection in project: deassign repos from a project

txels:
Now one can also deassign repos from a project.


alexcouper:
Works well :)

PR #80, opened 2014-11-30
List branches for a project, with filtering

txels:
A new activity tab for a project, shows branches in that project's repos, with filtering.

### DIFF
--- a/ployst/core/repos/views.py
+++ b/ployst/core/repos/views.py
@@ -34,4 +34,4 @@ class RepositoryViewSet(PermissionsViewSetMixin, ModelViewSet):
 
 class BranchViewSet(PermissionsViewSetMixin, ModelViewSet):
     model = Branch
-    filter_fields = ('name', 'repo')
+    filter_fields = ('name', 'repo', 'repo__project')
--- a/ployst/ui/static/repos/project-branches.html
+++ b/ployst/ui/static/repos/project-branches.html
@@ -1,9 +1,11 @@
-<p>Filter: <input ng-model="filter" type="text"></p>
-
 <table class="table">
     <thead>
         <tr>
-            <th>Branch Name</th>
+            <th>Branch Name &nbsp;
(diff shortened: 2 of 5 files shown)


ployst/ui/static/repos/project-branches.html
```diff
@@ -0,0 +1,22 @@
...
+    <thead>
+        <tr>
+            <th>Branch Name &nbsp;
+                <small>
+                    <input ng-model="filter" type="text">
+                </small>
+            </th>
+            <th>Head Commit</th>
+        </tr>
+    </thead>
```
alexcouper, line 12: I think we should show repo here too

txels, line 12: Yes. I think so too.

For this case, I think I'll retrieve from the repos endpoint, that returns nested branches.

But while trying to do this one and the previous story "quickly and naively" (trying to use existing API only, if possible, in order to focus solely on the UI bits) I found myself a bit uncomfortable with the current API design - I think it does not really reflect resource structure (repos inside projects etc, as it would if we e.g. did `/project/123/repos` instead of `/repo?project=123` as we do now). Maybe something to chat about.

txels:
@alexcouper done.

ployst/core/repos/views.py
```diff
@@ -34,4 +34,4 @@ def create(self, request, *args, **kwargs):
 
 class BranchViewSet(PermissionsViewSetMixin, ModelViewSet):
     model = Branch
-    filter_fields = ('name', 'repo')
+    filter_fields = ('name', 'repo', 'repo__project')
```
txels, line 37: Now we're not using this endpoint anymore.

ployst/ui/static/repos/project-branches.html
alexcouper, line 12: Yep sounds worthy of a refactor.

PR #82, opened 2014-12-02
Tweaks to make the GH hook work against the core repo API

txels:
I believe this got lost in the models refactor where we minimised some models and got rid of teams.
I added a docstring to the test to track why it's there.

### DIFF
--- a/ployst/core/repos/test/test_api.py
+++ b/ployst/core/repos/test/test_api.py
@@ -29,6 +29,26 @@ class TestFiltering(ProjectTestMixin, APITestCase):
         self.assertEquals(len(repos), 1)
         self.assertEquals(repos[0]['name'], repo1.name)
 
+    def test_get_repos_by_owner_and_name(self):
+        """
+        Search by owner and name to get a list of matching repos.
+
+        This endpoint is used by the github hook. Do not remove this test
+        unless the hook no longer uses this.
+        """
+        repo1 = RepositoryFactory(owner='me', name='A', project=self.project)
+        RepositoryFactory(owner='you', name='B', project=self.project)
+        url = reverse('core:repos:repository-list')
+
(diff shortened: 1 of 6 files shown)


ployst/github/test/__init__.py
```diff
@@ -80,7 +80,7 @@ def ensure_dummy_clone_available():
         os.system('git clone git://github.com/pretenders/dummyrepo.git {0}'
                   .format(DUMMY_REPO))
     else:
-        cmd = 'git --git-dir="{0}/.git" fetch'.format(DUMMY_REPO)
+        cmd = 'git --git-dir="{0}/.git" fetch -p'.format(DUMMY_REPO)
```
txels, line 83: I was playing with the hook by pushing to a new branch, and a git repo test failed.
Deleted the branch on remote, and it still failed, so I realised this was missing.

ployst/core/repos/test/test_api.py
```diff
@@ -29,6 +29,26 @@ def test_get_repos_by_project(self):
         self.assertEquals(len(repos), 1)
         self.assertEquals(repos[0]['name'], repo1.name)
 
+    def test_get_repos_by_owner_and_name(self):
+        """
+        Search by owner and name to get a list of matching repos.
+
+        This endpoint is used by the github hook. Do not remove this test
+        unless the hook no longer uses this.
```
txels, line 37: Maybe these type of tests (API used by another service) should live in a special place, or we could use a special `@attr`, instead of just a docstring?

alexcouper, line 37: Yes this is a good idea I think. At least then we can double check for breaking changes.

ployst/github/test/__init__.py
alexcouper, line 83: Makes sense that we want to do this.

PR #84, opened 2014-12-03
83 store branches

alexcouper:
This solves #83 

I've tested and I do indeed receive new branches in the UI (upon refresh) following a github post :)

PR #86, opened 2014-12-03
Get head revision of a branch on github hook

txels:
Updates #84
Turns out we can easily get commit ID from the web hook.

### DIFF
--- a/ployst/github/views/hook.py
+++ b/ployst/github/views/hook.py
@@ -14,6 +14,15 @@ from ..tasks.hierarchy import recalculate, update_branch_information
 LOGGER = logging.getLogger(__name__)
 
 
+def safe_get(payload, keys, default=''):
+    try:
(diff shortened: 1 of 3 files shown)


ployst/github/views/hook.py
```diff
@@ -39,7 +48,8 @@ def receive(request):
         if settings.GITHUB_CALCULATE_HIERARCHIES_ON_HOOK:
             recalculate.delay(org, repo, branch_name)
         else:
-            update_branch_information.delay(org, repo, branch_name)
+            head = safe_get(commit_info, ['head_commit', 'id'])[:7]
```
txels, line 51: We also can get the name of the commit author, which we could add to the branch model, if we think it's worthy info to show.

alexcouper:
Manually ran tests as travis failed to install reqs. Merging.

PR #90, opened 2014-12-04
Towards a less proprietary UI

### LINKED ISSUE #81: UI fun
I think we may be getting rid of the left hand bar, but anyway, there's some weirdness going on when you scroll right with a smaller window:

screen shot 2014-12-01 at 10 02 42 pm

txels:
I believe this fixes https://github.com/pretenders/ployst/issues/81 but the main point is to get rid of proprietary UI components and use stock, well-tested ones.

I've also refactored the way we inject Django-based URLs into angular.

### DIFF
--- a/ployst/templates/base-angular.html
+++ b/ployst/templates/base-angular.html
@@ -19,16 +19,11 @@
 <script src="{{ STATIC_URL }}lib/angular-route/angular-route.min.js"></script>
 <script src="{{ STATIC_URL }}lib/angular-cookies/angular-cookies.min.js"></script>
 <script type="text/javascript">
-    var STATIC_URL = "{{ STATIC_URL }}";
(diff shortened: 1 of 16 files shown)


ployst/templates/base-angular.html
```diff
@@ -11,12 +11,24 @@
...
-    window.STATIC_URL = {{ STATIC_URL }};
+    var STATIC_URL = "{{ STATIC_URL }}";
+
+    angular.module('ployst.django', [
+    ])
+    .constant('Django', {
+        URL: {
+            STATIC: STATIC_URL,
+            HOME: "{% url 'ui:home' %}",
+            LOGOUT: "{% url 'logout' %}"
```
txels: This is a way to make reversed URLs available to angular apps.

PR #92, opened 2014-12-05
Separate project.html into separate directive

txels:
Splits up projects and project.

### DIFF
--- a/ployst/ui/static/app.js
+++ b/ployst/ui/static/app.js
@@ -39,7 +39,7 @@ var ployst = angular.module('ployst', [
                 })
                 .when('/projects', {
                     controller: 'projects',
-                    templateUrl: Django.URL.STATIC + 'projects/project.html',
+                    templateUrl: Django.URL.STATIC + 'projects/projects.html',
                     menu: 'projects'
                 })
                 .otherwise({
--- a/ployst/ui/static/projects/ployst.projects.js
+++ b/ployst/ui/static/projects/ployst.projects.js
@@ -104,6 +104,16 @@ angular.module('ployst.projects', [
             };
         }
     ])
+    .directive('project', [
+        'Django',
+
+        function(Django) {
+            return {
+                restrict: 'E',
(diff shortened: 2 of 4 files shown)


ployst/ui/static/app.js
```diff
@@ -39,7 +39,7 @@ var ployst = angular.module('ployst', [
                 })
                 .when('/projects', {
                     controller: 'projects',
-                    templateUrl: Django.URL.STATIC + 'projects/project.html',
+                    templateUrl: Django.URL.STATIC + 'projects/projects.html',
```
txels, line 42: I renamed this file incorrectly a few PRs ago.

ployst/ui/static/projects/ployst.projects.js
```diff
@@ -104,6 +104,16 @@ angular.module('ployst.projects', [
             };
         }
     ])
+    .directive('project', [
+        'Django',
+
+        function(Django) {
+            return {
+                restrict: 'E',
```
alexcouper, line 112: What does this restrict do?

ployst/ui/static/projects/ployst.projects.js
```diff
@@ -104,6 +104,16 @@ angular.module('ployst.projects', [
             };
         }
     ])
+    .directive('project', [
```
alexcouper, line 107: Is this how the magic works? As in the name of the html field matches this and that's why it all works?

txels, line 107: Yes this is the "tag" name (because this directive is used as a tag). Angular does some renaming, e.g. camel case to dash separators. `'projectsMenu'` becomes tag `<projects-menu>`.

Ha yes this may explain things :)

ployst/ui/static/projects/ployst.projects.js
txels, line 112: `E`: element (tag). This directive can only be used as a tag name. Other options are `A`ttribute (`<div project>`) or class attribute etc... I in general prefer using tags only - even if that doesn't work on old explorer versions.

PR #93, opened 2014-12-05
Sidebar finally working as intended

txels:
I got rid of the proprietary component, and borrowed from a sidebar example (my CSS is very limited).
It works much better, and with less code.

Responsive-wise, we're not yet there, but this gets us a bit closer.


alexcouper:
Super shiny!

PR #102, opened 2015-01-23
Generalise hook handling, in preparation to handle PRs

txels:
Part of task https://txels.tpondemand.com/entity/365

This is an initial refactor to generalise how we handle github hooks. I'd like to actually extract the module `hooker.py` into a separate installable library e.g. "github-hooker".

I haven't written specific unit tests for the new module, but all existing tests pass, and the github_poke command works. If we decide to pull this out as a library, I'll move/rewrite tests accordingly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ tags
 .ropeproject/
 /assets/lib
 /node_modules/
+github-hook*.json
 
 # Editors and IDEs
 .project
@@ -22,9 +23,9 @@ tags
 /docs/build
 /ployst.egg-info
 /static
+/.tox/
 
 # Some OSX specific thing
 *.DS_Store
-/.tox/
 
 ployst/github/test/data/
--- a/ployst/github/hooker.py
+++ b/ployst/github/hooker.py
@@ -1,0 +1,87 @@
+from datetime import datetime
+import hashlib
+import hmac
+import json
+
+
+DEBUG = True
+
+
+class GithubHookException(Exception):
+    pass
+
+
+def compute_github_signature(body, secret):
+    return hmac.new(secret, body, hashlib.sha1).hexdigest()
+
+
+class GithubHookHandler(object):
+    """
+    A general purpose-class to help build github hook handlers.
+
+    See https://developer.github.com/webhooks/ for Github's webhook spec.
+
+    NOTE: This module may be spin-off into a separate library.
+    """
+
+    @staticmethod
+    def get_org_repo(request):
+        try:
+            payload = json.loads(request.body)
+            url = payload['repository']['url']
+            org, repo = url.split('/')[-2:]
+            return org, repo
+        except ValueError:
+            raise GithubHookException("Invalid payload")
+
+    def __init__(self, secret, request):
+        hub_sig = request.META['HTTP_X_HUB_SIGNATURE']
+        payload = request.body
+
+        if not self.validate_hook_post(payload, secret, hub_sig):
+            raise GithubHookException("Incorrect signature")
+
+        self.payload = json.loads(request.body)
+        self.event = request.META['HTTP_X_GITHUB_EVENT']
+
+        if DEBUG:
+            filename = 'github-hook-{}.json'.format(datetime.now().isoformat())
+            with open(filename, 'w') as f:
+                f.write(request.body)
+
+    @property
+    def org_repo(self):
+        url = self.payload['repository']['url']
+        org, repo = url.split('/')[-2:]
+        return org, repo
+
+    def validate_hook_post(self, body, secret, hub_signature):
+        """
+        Check that the post is legitimate.
+
+        A post is considered valid if the HMAC digest of the body equals the
+        header given in `X-Hub-Signature`.
+
+        See https://developer.github.com/v3/repos/hooks/#example
+        """
+        computed = compute_github_signature(body, secret)
+        sent_sig = hub_signature.split('sha1=')[-1]
+        return computed == sent_sig
+
+    def route(self):
+        """
+        Routes to the correct method based on github event.
+
+        It will call a method named `on_<event>` and provide the specific
+        part of the payload that corresponds to the event name.
+
+        See https://developer.github.com/v3/activity/events/types/ for details.
+
+        """
+        # Ensure self.org and self.repo are initialised before the handler
+        # method is called
+        self.org, self.repo = self.org_repo
+
+        handler_name = 'on_{}'.format(self.event)
+        if hasattr(self, handler_name):
+            getattr(self, handler_name)(self.payload)
--- a/ployst/github/management/commands/github_poke.py
+++ b/ployst/github/management/commands/github_poke.py
@@ -6,7 +6,8 @@ import requests
 from django.core.management.base import BaseCommand, CommandError
 from django.core.urlresolvers import reverse
 
-from ...github_client import compute_github_signature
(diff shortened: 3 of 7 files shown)


.gitignore
```diff
@@ -22,9 +23,9 @@ tags
 /docs/build
 /ployst.egg-info
 /static
+/.tox/
 
 # Some OSX specific thing
 *.DS_Store
-/.tox/
 
 ployst/github/test/data/
```
txels, line 31: I'm not sure why we ignore this. There are some files here under version control.
Do you know @alexcouper ?

ployst/github/hooker.py
```diff
@@ -0,0 +1,87 @@
...
+        if not self.validate_hook_post(payload, secret, hub_sig):
+            raise GithubHookException("Incorrect signature")
+
+        self.payload = json.loads(request.body)
+        self.event = request.META['HTTP_X_GITHUB_EVENT']
+
+        if DEBUG:
+            filename = 'github-hook-{}.json'.format(datetime.now().isoformat())
+            with open(filename, 'w') as f:
+                f.write(request.body)
```
txels, line 50: This is something that I'm adding temporarily as a way to collect payload samples from all github activity.
It may be handy to keep something like this in the future?

ployst/github/hooker.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    A general purpose-class to help build github hook handlers.
+
+    See https://developer.github.com/webhooks/ for Github's webhook spec.
+
+    NOTE: This module may be spin-off into a separate library.
+    """
+
+    @staticmethod
+    def get_org_repo(request):
```
txels, line 28: I don't like this method much, and the partial duplication with `org_repo` below. The reason for this to exist is that we generate github secrets tailored to each repo (they include org and repo name). 

I'm not sure what the real reason behind that is, but besides the currently ugly code it poses one additional potential issue: in the future we may want to support organisation webhooks, like repo creation events (https://developer.github.com/v3/activity/events/types/#repositoryevent) and let ployst know when you have created a new repo, e.g. to suggest you to include it in a project.

What do you think @alexcouper ?

ployst/github/management/commands/github_poke.py
```diff
@@ -37,12 +38,19 @@ def handle(self, *args, **options):
...
+            'head_commit': {
+                'id': '1234567'
+            }
         })
-        sig = compute_github_signature(load, org, repo)
+        secret = get_secret(org, repo)
+        sig = compute_github_signature(load, secret)
 
         response = requests.post(full_url, data=load,
-                                 headers={'X_HUB_SIGNATURE': sig})
```
txels, line 45: For some reason, this header format didn't work for me...

ployst/github/views/hook.py
```diff
@@ -23,49 +23,52 @@ def safe_get(payload, keys, default=''):
...
-            update_branch_information.delay(org, repo, branch_name, head)
+            update_branch_information.delay(
+                self.org, self.repo, branch_name, head
+            )
 
-    return HttpResponse("OK")
 
+@csrf_exempt
+@require_http_methods(['POST'])
+def receive(request):
```
txels, line 56: Now the view itself has become really small, the way I like them :)

.gitignore
alexcouper, line 31: Might be because we clone the pretenders/dummy repo there in some tests.

Perhaps the path should be extended to `ployst/github/test/data/pretenders`

ployst/github/hooker.py
alexcouper, line 28: Isn't the reason for this secret being unique so that you can't fake requests to the API.
If we had one globally used one, you would only need to sign up for your repo to then use it to make subsequent requests pretending to be from github.

To solve your use case too, perhaps we should only use the organization name to generate the secret, rather than repo too?

alexcouper, line 28: But I agree the name and placement of this is not ideal...

ployst/github/management/commands/github_poke.py
alexcouper, line 45: Possibly some change in the recent django upgrade

ployst/github/test/test_end_to_end.py
```diff
@@ -15,7 +15,7 @@ class TestEndToEnd(TestCase):
     def setUp(self):
         ensure_dummy_clone_available()
 
-    @patch(__name__ + '.tasks.hierarchy.client', MockClient())
+    @patch('ployst.github.tasks.hierarchy.client', MockClient())
```
alexcouper, line 18: I had patched this like this so that github worked as a standalone app (rather than being reliant on being inside ployst).

I don't mind us losing that though.

ployst/github/hooker.py
```diff
@@ -0,0 +1,87 @@
...
+        A post is considered valid if the HMAC digest of the body equals the
+        header given in `X-Hub-Signature`.
+
+        See https://developer.github.com/v3/repos/hooks/#example
+        """
+        computed = compute_github_signature(body, secret)
+        sent_sig = hub_signature.split('sha1=')[-1]
+        return computed == sent_sig
+
+    def route(self):
```
alexcouper, line 71: this is nice

ployst/github/views/hook.py
```diff
@@ -23,49 +23,52 @@ def safe_get(payload, keys, default=''):
...
+            return HttpResponseBadRequest()
 
-    if branch_name:
         if settings.GITHUB_CALCULATE_HIERARCHIES_ON_HOOK:
-            recalculate.delay(org, repo, branch_name)
+            recalculate.delay(self.org, self.repo, branch_name)
         else:
-            head = safe_get(commit_info, ['head_commit', 'id'])[:7]
-            update_branch_information.delay(org, repo, branch_name, head)
+            update_branch_information.delay(
```
alexcouper, line 49: looks like we do already have celery stuff...

.gitignore
txels, line 31: Maybe we need the generated files to live in a different root folder altogether? I don't like mixing up folders with VC'd and non-VC'd stuff

alexcouper:
This is very tidy indeed.

Merging

ployst/github/hooker.py
txels, line 50: BTW I temporarily enabled "all" events on the hook and I'm collecting JSON files of all comments in this PR.

ployst/github/hooker.py
txels, line 28: Yes the issue is that we need to parse the payload first and check whether the request is legit afterwards... meaning we can't do both cleanly without duplication in this class, that I want to publish as a separate library. Maybe the generation of a secret per organisation could be part of this library as well, and be _standard behaviour_?

ployst/github/management/commands/github_poke.py
txels, line 45: Hrm no I think this is purely `requests` code. The header was not coming through. I don't think django was filtering it out, although it might have been.

ployst/github/test/test_end_to_end.py
txels, line 18: `mock` was failing to patch this. I'm also confused as to how it worked, since `__name__` will be the name of this test module, not the source one...?

alexcouper, line 18: I import `tasks` at the top of the module.

http://stackoverflow.com/questions/14341689/how-to-mock-using-patch-relative-paths

PR #110, opened 2015-03-20
Ensure latest ployst is runnable on Heroku

txels:
The github integration didn't work, as the GH provider needed to talk to core and find out what port it was running on.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: gunicorn ployst.wsgi
+web: gunicorn --workers 4 -k eventlet ployst.wsgi
--- a/docs/heroku-bootstrap.md
+++ b/docs/heroku-bootstrap.md
@@ -1,0 +1,31 @@
+# Bootstrapping ployst in Heroku
+
+## In Github
+
+Create an application entry in Github.
+
+    https://github.com/settings/applications/
+
+Callback URL: http://ployst.com/github/oauth-confirmed/
+
+Copy client ID and client secret to use as `GITHUB_CLIENT_ID` and
+`GITHUB_CLIENT_SECRET`.
+
+## In ployst
+
+Go to admin and create an API token, label github.
+Copy the token to use as `GITHUB_CORE_API_TOKEN`
+
+## In Heroku
+
+Go to ployst settings page:
+
+    https://dashboard.heroku.com/apps/ployst/settings
+
+Reveal config vars. Add/set values for the following:
+
+    GITHUB_CORE_API_TOKEN
+    GITHUB_CLIENT_ID
+    GITHUB_CLIENT_SECRET
+
+All these can also be set via `heroku config`.
--- a/ployst/settings/heroku.py
+++ b/ployst/settings/heroku.py
@@ -1,4 +1,7 @@
+from os import getenv
+
 from .base import *  # noqa
+
 
 # Parse database configuration from $DATABASE_URL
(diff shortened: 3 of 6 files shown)


ployst/ui/views.py
```diff
@@ -14,6 +15,7 @@ def get_context_data(self, *args, **kwargs):
         providers = get_all_providers_meta()
         data.update({
             'providers': providers,
+            'core_url': settings.CORE_API_ADDRESS
```
txels: I added this for debugging purposes, but it could either be deleted, generalised, or left for `DEBUG=True`, or replaced with debug toolbar.

Opinions?

docs/heroku-bootstrap.md
```diff
@@ -0,0 +1,31 @@
...
+
+    https://dashboard.heroku.com/apps/ployst/settings
+
+Reveal config vars. Add/set values for the following:
+
+    GITHUB_CORE_API_TOKEN
+    GITHUB_CLIENT_ID
+    GITHUB_CLIENT_SECRET
+
+All these can also be set via `heroku config`.
```
txels, line 31: Some of these instructions also apply to just getting ployst up and running locally...

ployst/settings/heroku.py
```diff
@@ -17,3 +20,14 @@
 # http://django-storages.readthedocs.org/en/latest/backends/amazon-S3.html
 
 DEBUG = True
+
+# We need to figure out or set the proper localhost port for gunicorn
+# CORE_API_ADDRESS = "http://localhost:<port>/"
+CORE_API_ADDRESS = "http://localhost:{}/".format(getenv('PORT', 8000))
```
txels: The `PORT` env var is set by heroku.

Procfile
```diff
@@ -1 +1 @@
-web: gunicorn ployst.wsgi
+web: gunicorn --workers 4 -k eventlet ployst.wsgi
```
txels, line 1: Else it would block when GH provider tried to call core via HTTP, as they are sharing the same gunicorn ATM.

alexcouper, line 1: Ah ok

ployst/ui/views.py
alexcouper: How does this help with debugging?

alexcouper:
Happy to merge as is

ployst/ui/views.py
txels: In particular when debugging why on Heroku the github provider hang up while talking to core. I wanted to make sure the port core was listening to was the one Heroku reported in the startup log.

I will remove this before merging. I'd rather later add debug toolbar.

Procfile
txels, line 1: Maybe we need to start running GH provider and core as separate processes.

ployst/ui/views.py
txels: Gone with the wind.

txels:
Happy for a merge on green.

txels:
@alexcouper merge?

txels:
I have a branch with Django 1.8 waiting for this to be merged.... (https://github.com/pretenders/ployst/compare/carles/heroku...carles/django-1.8?expand=1)

Procfile
alexcouper, line 1: I think the only reason we're not is due to heroku deployment and the need
to pay above 1 process?

PR #111, opened 2015-04-02
Keep django version up to date: 1.8 is out

txels:
This also required updating django rest framework and django-filter


alexcouper:
Well played, Sir.

txels:
Yeah let's stay up to date now that it's relatively cheap.

